PR #15, opened 2026-04-15
Replace per-type Positive/NonNegative numerics with generic Positive<T> and NonNegative<T>

KaliCZ:
## Summary

- Introduce **`Positive<T>`** and **`NonNegative<T>`** (`readonly struct where T : INumber<T>`) replacing the 8 hand-rolled per-type structs (`PositiveInt`, `PositiveDecimal`, `PositiveLong`, `PositiveShort` and their `NonNegative` counterparts).
- Follow the `TryCreate`/`Create` factory pattern (private ctor, `TryCreate` returns nullable, `Create` throws).
- `Positive<T>` stores `Value - T.One` internally so `default(Positive<T>).Value == T.One` — the zero-initialized struct satisfies the positivity invariant. Trade-off: one add on read; `T.One` JIT-folds for concrete instantiations so the cost is a single ALU op.
- `NonNegative<T>` stores `Value` directly — its `default` is `T.Zero`, already valid.
- Add checked-arithmetic `Sum` extensions for both types (`OverflowException` on overflow).
- Delete the 8 `_Old` type files + 8 `_Old` test files (–1477 lines).
- Minimally patch 3 remaining `_Old` files that referenced deleted types:
  - `NumberExtensions_Old.cs` — drop `AsPositive`/`AsNonNegative` sections
  - `IEnumerableExtensions_Numeric_Old.cs` — drop `Sum`/`SafeSum` for deleted types
  - `IEnumerableExtensions_ReadOnlyList_Old.cs` — drop `ElementAtOption(NonNegativeInt)` overload
- Remaining `Numeric_Old` files (`NonPositive*`, `Digit*`, `Divide`) are untouched — out of scope.

## Test plan

- [x] `dotnet build` — 0 errors, 0 new warnings
- [x] `dotnet test` — 416/416 passing
- [x] FsCheck property tests cover `int`, `long`, `decimal`, `short` for both types
- [x] Overflow (`checked` arithmetic) tested for `int` and `long`
- [x] Default-value tests: `default(Positive<int>).Value == 1`, equals `Create(1)`
- [x] `MaxValue` round-trip tested
- [x] Empty `Sum` for `Positive<T>` throws, for `NonNegative<T>` returns zero


KaliCZ:
I will review later - now merging this, so I can continue with future changes via cloud claude sessions that will built on this.

PR #22, opened 2026-04-16
Add numeric strong types with JSON/EF Core converters and integration tests

### LINKED ISSUE #18: Add JSON and EF Core value converters for numeric strong types
## Context

PR #15 introduces `Positive<T>` and `NonNegative<T>` generic numeric strong types. A follow-up (see linked issue) will add `Negative<T>` and `NonPositive<T>`. All four types need serialization and persistence support.

### JSON (`System.Text.Json`)

A `JsonConverterFactory` that produces a `JsonConverter<Wrapper<T>>` for each closed generic — e.g. `Positive<int>`, `NonNegative<decimal>`, etc. The converter should:

- **Serialize** by writing the underlying `T` value directly (no wrapper object on the wire).
- **Deserialize** by reading `T` and passing it through `Create`, so invalid values are rejected at the deserialization boundary.
- Register via `[JsonConverter]` attribute on the struct and/or via `JsonSerializerOptions.Converters`.

### EF Core (SQL Server + PostgreSQL)

A generic `ValueConverter<Wrapper<T>, T>` (or converter factory) so EF Core maps the column as the underlying `T`. Should work for both SQL Server and PostgreSQL (already tested via Testcontainers in `StrongTypes.Api`).

### Types to cover

| Type | Status |
|------|--------|
| `Positive<T>` | Exists (PR #15) |
| `NonNegative<T>` | Exists (PR #15) |
| `Negative<T>` | Planned (see linked issue) |
| `NonPositive<T>` | Planned (see linked issue) |

The converter factories should handle all four once they exist. Design them so adding a new numeric wrapper is a one-line registration, not a new converter class.

## Existing patterns to follow

- `NonEmptyStringJsonConverter` in `src/StrongTypes/Strings/` for the JSON side.
- `StrongTypes.Api` integration tests for the EF Core wire-to-DB round-trip pattern.

### LINKED ISSUE #19: Add Negative<T> and NonPositive<T> generic numeric types
## Context

PR #15 introduced `Positive<T>` and `NonNegative<T>` (`readonly struct where T : INumber<T>`). The old `Numeric_Old/` folder still contains per-type `NonPositiveInt`, `NonPositiveDecimal`, `NonPositiveLong`, `NonPositiveShort` that should be collapsed the same way.

Additionally, a `Negative<T>` type (strictly less than zero) is the symmetric counterpart to `Positive<T>` and completes the set.

## Types to add

| Type | Invariant | `default` value |
|------|-----------|-----------------|
| `Negative<T>` | `value < T.Zero` | `-T.One` (offset storage, same trick as `Positive<T>`) |
| `NonPositive<T>` | `value <= T.Zero` | `T.Zero` (valid as-is, same approach as `NonNegative<T>`) |

## Requirements

- `TryCreate` / `Create` factory pattern (private ctor), matching `Positive<T>` and `NonNegative<T>`.
- `IEquatable`, `IComparable`, implicit cast to `T`, explicit cast from `T`.
- `checked`-arithmetic `Sum` extension for both types.
- FsCheck property tests covering `int`, `long`, `decimal`, `short`.
- Delete the 4 `NonPositive*_Old` type files + 4 test files.
- Patch any remaining `_Old` files that reference the deleted types (same minimal-patch approach as PR #15).

## Existing patterns to follow

- `src/StrongTypes/Numbers/Positive.cs` and `NonNegative.cs` for the struct shape.
- `src/StrongTypes/Numbers/PositiveExtensions.cs` for the `Sum` extension.
- `src/StrongTypes.Tests/Numbers/` for test structure.

## Note

The JSON / EF Core converters for all four numeric types are tracked in a separate issue.

### LINKED ISSUE #20: Add IEquatable<T> and IComparable<T> to Positive<T> and NonNegative<T>
## Context

`NonEmptyString` implements both `IEquatable<NonEmptyString>` + `IEquatable<string>` and `IComparable<NonEmptyString>` + `IComparable<string>`. The numeric strong types introduced in PR #15 only implement the self-typed interfaces:

```csharp
// Current
public readonly struct Positive<T> :
    IEquatable<Positive<T>>,
    IComparable<Positive<T>>,
    IComparable
    where T : INumber<T>
```

They're missing `IEquatable<T>` and `IComparable<T>` for the underlying numeric type.

## Requirements

Add to both `Positive<T>` and `NonNegative<T>` (and future `Negative<T>` / `NonPositive<T>` when they land):

- `IEquatable<T>` — enables `Positive<int>.Create(5).Equals(5)` without boxing.
- `IComparable<T>` — enables `Positive<int>.Create(5).CompareTo(3)` directly.
- `Equals(T)` and `CompareTo(T)` method implementations.
- `==` / `!=` operators between `Positive<T>` and `T` (both directions).
- `<` / `<=` / `>` / `>=` operators between `Positive<T>` and `T` (both directions).
- Update `IComparable.CompareTo(object?)` to accept bare `T` values (matching `NonEmptyString`'s pattern for `string`).
- Update `Equals(object?)` to accept bare `T` values.

## Files to change

- `src/StrongTypes/Numbers/Positive.cs`
- `src/StrongTypes/Numbers/NonNegative.cs`
- `src/StrongTypes.Tests/Numbers/PositiveTests.cs`
- `src/StrongTypes.Tests/Numbers/NonNegativeTests.cs`

## Pattern to follow

`src/StrongTypes/Strings/NonEmptyString.cs` — see its equality and comparison regions for the dual-interface pattern.

KaliCZ:
## Summary

Implements issues #18, #19, and #20:

- **#19** — Add `Negative<T>` and `NonPositive<T>` generic numeric types with `TryCreate`/`Create` factories, `Sum` extensions, FsCheck property tests. Deletes 4 `NonPositive*_Old` type files and 4 test files; patches `NumberExtensions_Old`.
- **#20** — Add `IEquatable<T>` and `IComparable<T>` to all four numeric types (`Positive<T>`, `NonNegative<T>`, `Negative<T>`, `NonPositive<T>`) with cross-type `==`/`!=`/`<`/`<=`/`>`/`>=` operators in both directions.
- **#18** — Add `NumericStrongTypeJsonConverterFactory` (single factory for all four types, registered via `[JsonConverter]` attribute) and `NumericStrongTypeValueConverter<TWrapper, T>` for EF Core. Full API infrastructure for value-type entities (`IValueEntity`, `ValueEntityControllerBase`, value request/response models) with entities, controllers, and integration tests (create/get/update) for all four types against both SQL Server and PostgreSQL.

Closes #18, closes #19, closes #20

## Test plan

- [ ] Unit tests: FsCheck property tests for `Negative<T>` and `NonPositive<T>` (TryCreate, Create, equality, comparison, cross-type operators) across `int`, `long`, `decimal`, `short`
- [ ] Unit tests: Cross-type `IEquatable<T>`/`IComparable<T>` property tests added to existing `PositiveTests` and `NonNegativeTests`
- [ ] Integration tests: Create/Get/Update for `Positive<int>`, `NonNegative<int>`, `Negative<int>`, `NonPositive<int>` entities against both SQL Server and PostgreSQL via Testcontainers
- [ ] Integration tests: JSON serialization round-trip (camelCase, null handling for nullable value-type properties)
- [ ] Build: verify no remaining references to deleted `NonPositive*_Old` types

https://claude.ai/code/session_01Y5f4zR7KhKyNCfxXtGBBuD


src/StrongTypes.Api.IntegrationTests/Infrastructure/ValueIntegrationTestBase.cs
```diff
@@ -0,0 +1,83 @@
...
+/// </summary>
+public abstract class ValueIntegrationTestBase<TEntity, T>(TestWebApplicationFactory factory) : IDisposable
+    where TEntity : class, IValueEntity<TEntity, T>
+    where T : struct
+{
+    private readonly IServiceScope _scope = factory.Services.CreateScope();
+
+    protected HttpClient Client { get; } = factory.CreateClient();
+
+    protected SqlServerDbContext SqlDb => _scope.ServiceProvider.GetRequiredService<SqlServerDbContext>();
```
KaliCZ: A lot of this is shared with the reference type test class. Can we unify?

src/StrongTypes.Api.IntegrationTests/NumericItems/CreateNegativeIntEntityTests.cs
```diff
@@ -0,0 +1,49 @@
...
+    public async Task NonNullable_ZeroValue_ReturnsBadRequest()
+    {
+        var response = await Client.PostAsJsonAsync(NonNullable, Body(0, -1), Ct);
+        Assert.Equal(HttpStatusCode.BadRequest, response.StatusCode);
+    }
+
+    [Fact]
+    public async Task NonNullable_PositiveValue_ReturnsBadRequest()
+    {
+        var response = await Client.PostAsJsonAsync(NonNullable, Body(1, -1), Ct);
```
KaliCZ: The test cases need to validate errors not only for the first parameter, but also the second - the nullable one. Because sending a positive should crash and not just turn into null.


Also, please use attribute for defining the test cases - don't copy paste the test cases.  So for example there can be a single test that sends 2 requests and gets an array of invalid values as the test cases.

src/StrongTypes.Api.IntegrationTests/NumericItems/Negative/NegativeIntEntityTestBase.cs
```diff
@@ -0,0 +1,10 @@
+using StrongTypes.Api.Entities;
+using StrongTypes.Api.IntegrationTests.Infrastructure;
+
+namespace StrongTypes.Api.IntegrationTests.NumericItems;
+
+public abstract class NegativeIntEntityTestBase(TestWebApplicationFactory factory)
```
KaliCZ: We probably need folders per type at this point. With the actual shared code being just in the numeric folder.

src/StrongTypes.Api/Controllers/ValueEntityControllerBase.cs
```diff
@@ -0,0 +1,73 @@
...
+    public Task<IActionResult> UpdateNullable(Guid id, NullableValueRequest<T> request) =>
+        UpdateAsync(id, request.Value, request.NullableValue);
+
+    private async Task<IActionResult> GetAsync(DbSet<TEntity> set, Guid id)
+    {
+        var entity = await set.FindAsync(id);
+        return entity is null ? NotFound() : Ok(ValueEntityDto<T>.From(entity));
+    }
+
+    private async Task<IActionResult> CreateAsync(T value, T? nullableValue)
```
KaliCZ: A lot of this code is shared with the reference types - e.g everything about ef core - can we unify the code?

src/StrongTypes.Api/Entities/NegativeIntEntity.cs
```diff
@@ -0,0 +1,26 @@
...
+        Id = Guid.NewGuid();
+        Value = value;
+        NullableValue = nullableValue;
+    }
+
+    public Guid Id { get; set; }
+    public Negative<int> Value { get; set; }
+    public Negative<int>? NullableValue { get; set; }
+
+    public void Update(Negative<int> value, Negative<int>? nullableValue)
```
KaliCZ: Update can be implemented on the interface.

Not sure about the constructor. If yes, let's do it as well.

src/StrongTypes.Api/Infrastructure/NumericStrongTypeValueConverter.cs
```diff
@@ -0,0 +1,37 @@
...
+/// (<see cref="Positive{T}"/>, <see cref="NonNegative{T}"/>, <see cref="Negative{T}"/>,
+/// <see cref="NonPositive{T}"/>). Maps the wrapper to its underlying <typeparamref name="T"/>
+/// column via the <c>Value</c> property and <c>Create</c> factory.
+/// </summary>
+public sealed class NumericStrongTypeValueConverter<TWrapper, T> : ValueConverter<TWrapper, T>
+    where TWrapper : struct
+    where T : struct
+{
+    public NumericStrongTypeValueConverter()
+        : base(BuildToProvider(), BuildToModel())
```
KaliCZ: Will this benefit from caching the expression? If yes, let's do it.

src/StrongTypes.Tests/Numbers/NegativeTests.cs
```diff
@@ -0,0 +1,179 @@
...
+    {
+        var na = Negative<int>.TryCreate(a);
+        var nb = Negative<int>.TryCreate(b);
+        if (na is null || nb is null)
+        {
+            return;
+        }
+
+        Assert.Equal(a == b, na.Value == nb.Value);
+        Assert.Equal(a == b, na.Value.Equals(nb.Value));
```
KaliCZ, line 91: Also do object equals.

KaliCZ (commented):
I haven't read everything yet.

src/StrongTypes/Numbers/Negative.cs
```diff
@@ -0,0 +1,109 @@
...
+    public int CompareTo(T? other) => other is null ? 1 : Value.CompareTo(other);
+
+    int IComparable.CompareTo(object? obj) =>
+        obj switch
+        {
+            null => 1,
+            Negative<T> other => CompareTo(other),
+            T other => CompareTo(other),
+            _ => throw new ArgumentException($"Object must be of type {nameof(Negative<T>)} or {typeof(T).Name}.", nameof(obj))
+        };
```
KaliCZ, line 93: IS it expected that IComparable.CompareTo throws an exception? Shouldn't the default be returning 0? 

Also, should there be a default case where the other type is numeric, to try and default to some kind of numeric comparison? For example int 3 is higher than float 2.0 - I don't know how C# handles this

src/StrongTypes/Numbers/NegativeExtensions.cs
```diff
@@ -0,0 +1,25 @@
...
+    public static Negative<T> Sum<T>(this IEnumerable<Negative<T>> values) where T : INumber<T>
+    {
+        T sum = T.Zero;
+        foreach (var value in values)
+        {
+            sum = checked(sum + value.Value);
+        }
+        return Negative<T>.Create(sum);
+    }
+}
```
KaliCZ: Instead of having multiple extension files for each type - let's use feature slices here as well - so let's have SumExtensions and put all the methods into 1 file.

src/StrongTypes/Numbers/NumericStrongTypeJsonConverterFactory.cs
```diff
@@ -0,0 +1,73 @@
...
+    public override bool CanConvert(Type typeToConvert) =>
+        typeToConvert.IsGenericType
+        && SupportedDefinitions.Contains(typeToConvert.GetGenericTypeDefinition());
+
+    public override JsonConverter CreateConverter(Type typeToConvert, JsonSerializerOptions options)
+    {
+        var innerType = typeToConvert.GetGenericArguments()[0];
+        var converterType = typeof(Inner<,>).MakeGenericType(typeToConvert, innerType);
+        return (JsonConverter)Activator.CreateInstance(converterType)!;
+    }
```
KaliCZ: We could actually cache the instances, right? 
The converters can be singletons - they're not IDisposable, right?

src/StrongTypes/Numbers/Negative.cs
KaliCZ: The throw matches the .NET convention for `IComparable.CompareTo(object?)` — e.g. `int.CompareTo((object)"foo")` throws `ArgumentException`, it does not return 0. Returning 0 would assert equality across unrelated types, which is semantically wrong and breaks sort stability (`list.Sort()` would intermingle incomparable items).

For cross-numeric comparison (`Negative<int>` vs `float`), .NET built-ins also do not support it via the object-based `IComparable` path — `((int)3).CompareTo((object)(double)2.0)` throws. Widening conversions are only available through the typed path where the caller explicitly opts in by choosing the target type. I think matching that behavior is safer than silently converting (what if the comparison straddles `long.MaxValue` and `double`, where the round-trip is lossy?).

Happy to revisit if there is a concrete use case — we could add a typed `CompareTo(Positive<T> other)` overload so the sibling wrappers compare cleanly by `Value` without going through `object`. Let me know.

src/StrongTypes.Api/Infrastructure/StrongTypesModelConfigurationExtensions.cs
```diff
@@ -13,6 +13,16 @@ public static class StrongTypesModelConfigurationExtensions
...
         configurationBuilder.Properties<NonEmptyString>().HaveConversion<NonEmptyStringValueConverter>();
+
+        configurationBuilder.Properties<Positive<int>>()
+            .HaveConversion<NumericStrongTypeValueConverter<Positive<int>, int>>();
+        configurationBuilder.Properties<NonNegative<int>>()
+            .HaveConversion<NumericStrongTypeValueConverter<NonNegative<int>, int>>();
+        configurationBuilder.Properties<Negative<int>>()
+            .HaveConversion<NumericStrongTypeValueConverter<Negative<int>, int>>();
+        configurationBuilder.Properties<NonPositive<int>>()
+            .HaveConversion<NumericStrongTypeValueConverter<NonPositive<int>, int>>();
```
KaliCZ: We need to register the other types as well - long, decimal, single, double, short. Is there a way to do it in a nicer way than copy pasting all of them?

PR #38, opened 2026-04-19
Delete Option extensions and tests that nothing else depends on

KaliCZ:
Follow-up to #36. Strips everything Option-related that isn't actually load-bearing. `Option<T>` itself stays because `Try_Old` still exposes `Option<TSuccess>`/`Option<TError>` on its public surface and every `Coproduct_Old` property returns `Option<T>` — a full migration off `Option` has to wait until `Try` and `Coproduct` get rewritten.

## What got deleted

**Library:**
- `src/StrongTypes/Option_Old/IOptionExtensions_Old.cs` — the whole user-facing extension surface (`Map`/`FlatMap`/`Select`/`SelectMany`/`Where`/`Is`/`GetOr*`/`OrElse`/`Flatten`/`ToNullable`/`MapAsync`/`FlatMapAsync`/`MatchAsync`/`ToTry`).
- `src/StrongTypes/Option_Old/IOptionExtensions_CollectionsFlattening_Old.cs`.
- `src/StrongTypes/Option_Old/IOptionExtensions_CollectionsGetting_Old.cs` — whole file was already `[Obsolete]`.
- `src/StrongTypes/Extensions_Old/IReadOnlyDictionaryExtensions_Old.cs` — `.Get()` returning `Option<V>`; only consumer was `DictionaryTests_Old`.

**Library (trimmed in place):**
- `IEnumerableExtensions_Options_Old` keeps only `Flatten<IEnumerable<Option<T>>>` (used by `Try_Old.Aggregate`). `SafeMax`/`SafeMin`/`First`/`Last`/`SingleOption` on `IEnumerable<T>` gone — only _Old tests called them.
- `IEnumerableExtensions_ReadOnlyList_Old` loses `SingleOption`/`FirstOption`/`LastOption`. The non-Option helpers (`IsMultiple`, `Single`, `First`, `Second`…`Fifth`, `Last`, `ElementAt`, `IndexOf`) stay.
- `ObjectExtensions_Generic_Old` loses `.As<A>()` and both `ToOption<A>` overloads. `object as A` is the modern substitute for `As<A>`; `ToOption` is redundant given Maybe.
- `Option_Old.True`/`False`/`Unit` now use `Option.Valued(...)` directly instead of the deleted `value.ToOption()` extension.
- Two small `.GetOrDefault()`/`.ToOption()` call sites inside `_Old` (`IEnumerableExtensions_Emptiness_Old`, `IEnumerableExtensions_Try_Old`) now use `Option.Valued(...)` / `option.Value` directly.

**Tests:**
- `Options_Old/` in full (22 files).
- `Dictionary_Old/DictionaryTests_Old.cs` (tested a deleted extension).
- `Generative_Old/OptionGenerators_Old`, `OptionShrinkers_Old`, `Shrinkers_Old` (Shrinkers only exposed OptionShrinkers), `GeneratorHelpers_Old` (SometimesEmpty + unused ToList).
- `Collections_Old/First/Last/Single/FlattenTests_Old` and `INonEmptyEnumerable_Old/Option_CreateAndCreateFlatTests_Old`.
- `IEnumerableExtensionsTests_Old` drops the `FirstOption` case; `PartitionMatch` and `Aggregate` tests stay with `Option.Valued` replacing `.ToOption()`.
- `NonEmptyStringEqualityTests_Old` drops `OptionEqualityTest`; the non-Option `EqualityTest` stays as the only dedicated coverage of `NonEmptyString`'s equality operators.
- `OptionAssert_Old` kept but slimmed to three methods — still needed by `ExceptionsAggregateTests_Old`, `AsNonEmptyTests_Old`, `NonEmptyEnumerableTypeIssuesTests_Old` which assert over `Option<T>` results from non-deleted APIs.

## Numbers

- **4,700 lines deleted, 22 added.**
- Unit-test count: **716 → 365**, all still passing. The deleted tests exercised code that no longer exists.

## Interaction with #36

Independent. The #36 `maybe-type` branch removed the Option-returning `SafeMax`/`SafeMin` overloads from `IEnumerableExtensions_Options_Old` to clear the collision with Maybe's replacements — this PR deletes the same two methods (plus much more). Whichever merges first, the other rebases cleanly onto the result.

## Test plan

- [x] Full build clean (0 warnings, 0 errors).
- [x] Unit tests green (365/365).
- [ ] CI pipeline green (monitoring).


KaliCZ:
Closing for now — this cleanup only makes sense once Try and Coproduct have Result and OneOf replacements. Until then Option's extensions are borderline useless but the type itself still needs to stay as Try/Coproduct surface, which leaves an awkward middle state.

PR #43, opened 2026-04-20
Switch Maybe<T> JSON to flat merge-patch shape

KaliCZ:
## Summary

`Maybe<T>` previously serialized as `{"Value": x}` / `{"Value": null}`. The wrapper object was chosen to give `Maybe<T>?` tri-state PATCH semantics (absent = no-op, `{}` = clear, `{"Value": x}` = set), but that shape is non-idiomatic — RFC 7396 (JSON Merge Patch), RFC 6902 (JSON Patch), Protobuf FieldMask, and GraphQL all encode the same three intents through key presence on a flat value.

This PR moves the wire format to the flat shape:

| JSON          | `Maybe<T>?` value    | Intent               |
| ------------- | -------------------- | -------------------- |
| field omitted | `null`               | leave field untouched |
| `null`        | `Maybe<T>.None`      | clear field to `null` |
| `x`           | `Maybe<T>.Some(x)`   | set field to `x`      |

### Implementation

- `MaybeJsonConverter.cs` — `Maybe<T>` now writes `Some(x)` as the raw JSON of `x` and `None` as `null`. The factory also produces a dedicated `JsonConverter<Maybe<T>?>` with `HandleNull = true`, mapping JSON `null` to `Maybe<T>.None` and a value to `Maybe<T>.Some(x)`. STJ's built-in `Nullable<T>` handling would otherwise intercept `null` before our converter runs and collapse the *clear* case into a null reference.
- `StrongTypes.Api/Program.cs` — registers the factory in `JsonSerializerOptions.Converters`. The `[JsonConverter]` attribute on `Maybe<T>` alone isn't enough because attributes on `Maybe<T>` don't carry over to `Nullable<Maybe<T>>`.
- `readme.md` — the `### JSON` and `### Idiomatic usage: tri-state PATCH` sections document the new flat shape and the factory-registration requirement.
- `EntityModels.cs` — XML doc on `StructEntityPatchRequest` / `ReferenceEntityPatchRequest` updated to describe merge-patch semantics.

### Tests

- `MaybeJsonConverterTests.cs` rewritten for the flat shape. Full read/write matrix on both sides:
  - **Top-level `Maybe<T>`** — read null/value, write None/Some, reference-type, roundtrips, inner-converter interop (NonEmptyString, Positive\<int\>) including invalid-inner-value throws.
  - **Top-level `Maybe<T>?`** — every non-nullable top-level case mirrored: int, string, NonEmptyString, Positive\<int\>; read null ⇒ present None; write null/None/Some all emit `null`/`null`/`x`.
  - **Parent-record properties** — `PatchDto(int?, Maybe<int>?)` covers absent / null / value on both read and write; `PlainDto(Maybe<int>)` covers the flat-shape collapse (absent and null both ⇒ None) on read and write.
- `EntityTests.cs` PATCH suite updated to send the flat JSON (`{"nullableValue": x}`, `{"nullableValue": null}`, field absent). Dropped the now-redundant `Patch_NullableValueWithExplicitNullInner_ClearsNullableValue`. Semantics comment block at the top rewritten in merge-patch terms.

## Test plan

- [ ] `dotnet build` passes
- [ ] `dotnet test src/StrongTypes.Tests` passes (unit converter tests, including the new nullable parity set)
- [ ] `dotnet test src/StrongTypes.Api.IntegrationTests` passes (PATCH flat-JSON e2e across SQL Server + PostgreSQL)
- [ ] Skim the rendered readme to confirm the flat-PATCH table and the factory-registration example look right

https://claude.ai/code/session_01Tx7cofXtcsTkPbYyFTY5Pn

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -106,17 +106,17 @@ var missing =
 
 ### JSON
 
-`Maybe<T>` serializes via `System.Text.Json` as `{ "Value": x }` for `Some` and `{ "Value": null }` for `None`. Deserialization also accepts `{}` for `None`, so callers can omit the property entirely.
+`Maybe<T>` serializes via `System.Text.Json` to the raw JSON representation of the inner value: `Some(x)` writes as `x` and `None` writes as `null`. Deserialization is the inverse — `null` reads as `None`, any value reads as `Some`. The wire format matches what a plain `T?` would produce, so callers and consumers don't need to know the property is a `Maybe` at all.
 
 ### Idiomatic usage: tri-state PATCH
 
-HTTP `PATCH` has a long-standing modelling problem for nullable fields: a request needs to distinguish three intents — *don't touch this field*, *clear this field to null*, and *set it to a new value*. A plain `T?` collapses the first two cases. `Maybe<T>?` keeps them apart, because `Maybe<T>` itself is a value, so wrapping it in `T?` adds a real third state:
+HTTP `PATCH` has a long-standing modelling problem for nullable fields: a request needs to distinguish three intents — *don't touch this field*, *clear this field to null*, and *set it to a new value*. A plain `T?` collapses the first two cases, because on the wire "property absent" and "property set to null" both bind to a null reference. `Maybe<T>?` keeps them apart, because `Maybe<T>` itself is a value, so wrapping it in `Nullable<>` adds a real third state — and the tri-state is encoded on the wire using the standard JSON Merge Patch ([RFC 7396](https://datatracker.ietf.org/doc/html/rfc7396)) conventions:
 
-| JSON                     | Property value      | Intent                |
-| ------------------------ | ------------------- | --------------------- |
-| field omitted, or `null` | `null`              | leave field untouched |
-| `{}` or `{"Value":null}` | `Maybe<T>.None`     | clear field to `null` |
-| `{"Value":x}`            | `Maybe<T>.Some(x)`  | set field to `x`      |
+| JSON          | Property value       | Intent                |
+| ------------- | -------------------- | --------------------- |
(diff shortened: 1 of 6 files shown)


KaliCZ:
Closing this PR. The goal was to switch `Maybe<T>` JSON to a flat merge-patch shape where `Maybe<T>?` distinguishes absent / null / value on the wire, working out of the box with no converter registration. After exhausting the design space, that combination isn't reachable while keeping `Maybe<T>` a struct:

- `[JsonConverter]` on `Maybe<T>` is not propagated by STJ to `Nullable<Maybe<T>>`. The factory never sees the nullable type via attribute resolution.
- STJ's built-in `NullableConverterFactory` unconditionally wraps any nullable, and its `NullableConverter<T>` intercepts JSON `null` before the inner converter runs — `HandleNull = true` on the inner converter is ignored for nullables.
- Auto-registering a global converter (module initializer, `[assembly]` attributes, hosting startup, source generators, C# interceptors) either doesn't hook STJ's per-instance `JsonSerializerOptions` or requires consumers to use a specific options instance, i.e. opt-in by another name.
- Making `Maybe<T>` a class, or turning it into a three-state struct (`Undefined` / `None` / `Some`) both work technically but change the type's semantics in ways we don't want.

So the tri-state PATCH body shape will need an explicit opt-in (e.g. `options.Converters.Add(new MaybeJsonConverterFactory())`) in any future iteration, and the shape of that opt-in isn't something this PR should decide unilaterally.

I'll open a follow-up PR adding DTO-level JSON round-trip tests against the existing `{"Value": x}` converter shape, so we have concrete coverage of the request-body behavior we actually ship today.

Issue #42, opened 2026-04-20
Modernize NonEmptyEnumerable: JSON serialization + rework


KaliCZ:
`NonEmptyEnumerable<T>` / `INonEmptyEnumerable<T>` (currently under `src/StrongTypes/Collections_Old/`) is a useful primitive but predates the rest of the current StrongTypes conventions and is missing the pieces other types already have.

## Scope

- **JSON serialization** — add a `JsonConverter` so `INonEmptyEnumerable<T>` round-trips through `System.Text.Json` the same way other strong types do (see `NonEmptyStringJsonConverter`, `MaybeJsonConverter` for the pattern). Validate on deserialization: empty arrays should fail with a meaningful `JsonException`.
- **Factory pattern** — align `NonEmptyEnumerable.Create(...)` overloads with the `TryCreate` / `Create` pattern used elsewhere (nullable-returning `TryCreate` for the validated path, throwing `Create` as a thin wrapper).
- **Migrate out of `_Old`** — move the rewritten type and its extensions into a feature-sliced folder (`src/StrongTypes/Collections/` or similar). Follow the "drop `_Old` suffix on rework" rule in `CLAUDE.md`.
- **Modernize the implementation** — primary constructors, `#nullable enable`, minimize public surface, drop anything tied to the deleted `Option<T>` machinery.

## Out of scope

- EF Core value converter — can follow in its own PR once the JSON converter lands and the shape is stable.

KaliCZ:
Closed by #44.

PR #44, opened 2026-04-20
Modernize NonEmptyEnumerable: JSON converter, collection expressions, full extension surface

KaliCZ:
## Summary

Implements #42. Rewrites `NonEmptyEnumerable<T>` from the ground up: lifts it out of `Collections_Old/`, wires a `System.Text.Json` converter (for both the concrete type *and* `INonEmptyEnumerable<T>`), adds collection-expression support, `ICollection<T>` for LINQ fast paths, a struct `Enumerator`, vectorized equality, and a full extension surface — including **total** versions of LINQ's partial aggregators (`Max`, `Min`, `Aggregate`, `Average`, …) that never throw on non-empty input.

Also codifies two documentation rules in `CLAUDE.md` (XML comments are for callers only; `//` comments are not messages to future contributors) and applies them across the new files.

---

### Type + factories

- **Feature-sliced layout** — `NonEmptyEnumerable<T>`, `INonEmptyEnumerable<T>`, debug view, and the JSON converter live under `src/StrongTypes/Collections/`, replacing the `Option<T>`-coupled `Collections_Old/NonEmptyEnumerable_Old.cs`.
- **`Create(params ReadOnlySpan<T>)`** — variadic construction plus `[CollectionBuilder]` target for collection expressions: `NonEmptyEnumerable<int> list = [1, 2, 3];`. Empty `[]` throws at runtime.
- **`CreateRange(IEnumerable<T>?)`** / **`TryCreateRange(IEnumerable<T>?)`** — runtime-sequence factories; throw vs. return `null` for empty/null. `TryCreateRange` switches over `NonEmptyEnumerable<T>` / `T[]` / `List<T>` / `IReadOnlyList<T>` / `IReadOnlyCollection<T>` for single-allocation fast paths.
- **`ICollection<T>`** — `IsReadOnly = true`, real `Contains`/`CopyTo`, mutators throw. LINQ's `ToArray`/`Concat` recognize the type and pre-size.
- **Struct `Enumerator`** — allocation-free `foreach` on the concrete type.
- **Vectorized equality** via `MemoryExtensions.SequenceEqual<T>` (SIMD path for primitives / enums / Guid / unmanaged structs).
- **`AsSpan()`** for span-based consumers.

### JSON

- `NonEmptyEnumerableJsonConverterFactory` round-trips through STJ; attached via `[JsonConverter]` on both `NonEmptyEnumerable<T>` and `INonEmptyEnumerable<T>`, so interface-typed properties work without custom registration.
- Empty JSON array → `JsonException` at the boundary.
- Null elements are accepted (see behavior matrix below — consistent with plain `List<T>` under NRT).
- Round-trip tests are **string-anchored**: start from canonical JSON, deserialize, re-serialize, assert the bytes are identical. Catches canonical-form regressions, not just value equality.

### Extension surface

Extensions live in `NonEmptyEnumerableExtensions`. Where there's a meaningful span-based fast path (direct array access via `AsSpan`), we provide both a concrete overload on `NonEmptyEnumerable<T>` (compile-time dispatch) and an interface overload on `INonEmptyEnumerable<T>` (routes to the concrete one when possible, falls back to the indexer otherwise).

**Factory extensions**
- `source.AsNonEmpty()` — nullable return, mirrors the library-wide `AsX`/`ToX` convention.
- `source.ToNonEmpty()` — throws on empty/null.

**Invariant-preserving LINQ**
- `Select` / `Select(with index)` — span-based fast path.
- `SelectMany`, `Distinct`, `Flatten` — LINQ delegation (`ICollection<T>` already gives pre-sizing).
- `Concat(params ReadOnlySpan<T>)` — hand-rolled CopyTo; no LINQ equivalent for span input.
- `Concat(IEnumerable<T>)` — LINQ delegation.
- `Concat(this T head, params IEnumerable<T>[] tails)` — prepend a head to N tails. **Null tails throw `ArgumentNullException`** rather than being silently treated as empty (NRT discipline).
- `Reverse()` / `Prepend(T)` / `Append(T)`.
- `Take(Positive<int>)` — preserves non-emptiness at compile time.
- `Take(int)` — delegates via `ToPositive`; invalid counts fail loudly.

**Total aggregators** (non-emptiness eliminates the `InvalidOperationException` branch)
- `Max`, `Min`, `MaxBy`, `MinBy` — return `T`, not LINQ's pessimistic `TSource?`.
- `Last` — always defined, mirrors `Head`.
- `Aggregate(Func<T,T,T>)` (seedless) — total.
- `Average` — uses `checked` arithmetic so integer overflow throws `OverflowException` explicitly (float/double return `Infinity` — `checked` is a no-op for them).

### Integration tests — behavior matrix

New `CollectionJsonController` / `CollectionJsonTests` exercise the full `{IEnumerable<T>, IEnumerable<T?>, NonEmptyEnumerable<T>, NonEmptyEnumerable<T?>}` × `{int, Positive<int>, string, NonEmptyString}` matrix end-to-end through ASP.NET Core + STJ + EF Core.

| Input | `int` | `Positive<int>` | `string` | `NonEmptyString` |
|---|---|---|---|---|
| Valid array | ✅ round-trips | ✅ round-trips | ✅ round-trips | ✅ round-trips |
| Empty `IEnumerable<T>` | ✅ allowed | ✅ allowed | ✅ allowed | ✅ allowed |
| Empty `NonEmptyEnumerable<T>` | ❌ 400 | ❌ 400 | ❌ 400 | ❌ 400 |
| `null` element in nullable slot | ✅ kept | ✅ kept | ✅ kept | ✅ kept |
| `null` element in non-nullable slot | ❌ 400 (STJ) | ❌ 400 (struct rejects) | ⚠ **accepted** | ⚠ **accepted** |
| Invalid element (e.g. `0` for `Positive<int>`, `"   "` for `NonEmptyString`) | N/A | ❌ 400 | N/A | ❌ 400 |

The two ⚠ cells are cases where NRT annotations erase at runtime for reference `T`, so the converter can't distinguish `T` from `T?`. Consistent with what a plain `List<string>` would do. Documented explicitly in the tests and in the README callout.

### Documentation rules (`CLAUDE.md`)

- **XML comments are for callers only.** No implementation commentary, LINQ delegation notes, "why two overloads", etc. — those age badly and pollute IntelliSense.
- **`//` comments are not letters to future contributors.** Design rationale, history, "why isn't this simpler" explanations belong in the commit message / PR description, not in source.

Both rules applied across all new files (and retroactively across `NonEmptyEnumerable.cs` / `NonEmptyEnumerableJsonConverter.cs`).

### README

Rewrote the `NonEmptyEnumerable<T>` section: construction (Create, CreateRange, collection expressions), extensions (AsNonEmpty / ToNonEmpty), invariant-preserving LINQ chain, total-aggregator examples, `INonEmptyEnumerable<T>` covariance note with a `Dog → Animal` example, and JSON behavior (including the nullable-element caveat). Moved `Maybe<T>` under a new `Algebraic types` section.

---

## Test plan

- [x] `dotnet build` — 0 warnings, 0 errors.
- [x] `dotnet test src/StrongTypes.Tests` — **555 passed** (0 failed, 0 skipped).
- [x] `dotnet test src/StrongTypes.Api.IntegrationTests` — passes the full shape matrix on both SQL Server and PostgreSQL via Testcontainers.


src/StrongTypes/Collections_Old/IEnumerableExtensions_Concatenating_Old.cs
```diff
@@ -37,17 +37,6 @@ public static IEnumerable<T> SafeConcat<T>(this IEnumerable<T> first, params IEn
...
-    [Pure]
-    public static INonEmptyEnumerable<T> Concat<T>(this INonEmptyEnumerable<T> source, params T[] items)
-    {
-        return NonEmptyEnumerable.Create(source.Head, source.Tail.Concat(items).ToArray());
-    }
-
-    public static INonEmptyEnumerable<T> Concat<T>(this INonEmptyEnumerable<T> source, params IEnumerable<T>[] items)
-    {
-        return NonEmptyEnumerable.Create(source.Head, source.Tail.Concat(items).ToArray());
+        return NonEmptyEnumerable.CreateRange(others.Flatten().Prepend(e));
```
KaliCZ: This doesn't look like it should be faster. Is it?

PR #62, opened 2026-04-22
Add Result; retire Try

KaliCZ:
## Summary
- New `Result<T, TError>` and `Result<T>` (shorthand for `Result<T, Exception>`) replace the legacy `Try<A, E>`.
- Full API: `Map`/`MapAsync`, `MapError`/`MapErrorAsync`, `FlatMap`/`FlatMapAsync`, `FlatMapError`/`FlatMapErrorAsync`, bimap `Map<U, UError>(success, error)` + async, `Match`/`MatchAsync`, `Catch`/`CatchAsync` (plus typed `Catch<T, TException>`), `ThrowIfError`, `ToResult` from nullables (5 overloads per class/struct receiver), `Flatten`, `FlattenErrors`, `Partition`/`PartitionMatch`, `Aggregate` for arities 2–8 (tuple + combiner) and `IEnumerable`, equality (`==`/`!=`/`Equals(T)`/`Equals(TError)`).
- `Try` fully removed: `Try_Old`, the newer `Try` extensions folder, the `.tt` template, `Helpers.ttinclude`, and associated tests. Csproj `<None Update>` / `<Compile Update>` for the template are stripped — there was no NuGet dependency to remove (T4 was VS's built-in design-time generator).
- README: new `Result<T, TError>` section under Algebraic types with construction/inspection/transformation/exception-wrapping examples and an `Aggregate` validation example that composes a record via both implicit operators.

## Test plan
- [ ] Unit suite: `dotnet test src/StrongTypes.Tests` — 825 passing locally
- [ ] Integration suite still builds against both SQL Server and PostgreSQL harnesses
- [ ] Spot-check the README's `ParseUser` example compiles as written


readme.md
```diff
@@ -407,17 +406,102 @@ var missing =
...
+
+```csharp
+Result<int, string> r = Parse(input);
+
+if (r.Success is { } value) Console.WriteLine($"got {value}");
+if (r.Error   is { } msg)   Console.WriteLine($"failed: {msg}");
+```
+
+`IsSuccess` / `IsError` are there too when you don't need the payload.
+
```
KaliCZ, line 453: Also mention the ThrowIf methods

readme.md
```diff
@@ -407,17 +406,102 @@ var missing =
...
+`Map`, `MapError`, `Match`, and `FlatMap` let you chain without explicit branching:
+
+```csharp
+Result<int, string> r = Parse("42");
+
+// Map — transform the success value; errors pass through unchanged.
+Result<int, string> doubled = r.Map(x => x * 2);
+
+// Match — fold both branches into a single value.
+string message = r.Match(
```
KaliCZ, line 498: Can we implement support for  a C# switch statement? Then the match wouldn't need to exist.

readme.md
```diff
@@ -407,17 +406,102 @@ var missing =
...
+    //   T      → Result<T, TError>   (success branch)
+    //   TError → Result<T, TError>   (error branch)
+    // so both arms of the ternary flow into Result<T, string> without
+    // explicit wrapping.
+    Result<NonEmptyString, string> name =
+        NonEmptyString.TryCreate(nameInput) is { } n ? n : "name must not be empty";
+    Result<Positive<int>, string>  age  =
+        Positive<int>.TryCreate(ageInput)   is { } a ? a : "age must be positive";
+
+    return Result.Aggregate(name, age, (n, a) => new User(n, a))
```
KaliCZ: It seems that there should be an optional parameter to also map the errors into something in the aggregate method

readme.md
```diff
@@ -407,17 +406,102 @@ var missing =
...
+Result<User, string> ParseUser(string? nameInput, int ageInput)
+{
+    // Target-typed ternary on each line relies on two implicit operators:
+    //   T      → Result<T, TError>   (success branch)
+    //   TError → Result<T, TError>   (error branch)
+    // so both arms of the ternary flow into Result<T, string> without
+    // explicit wrapping.
+    Result<NonEmptyString, string> name =
+        NonEmptyString.TryCreate(nameInput) is { } n ? n : "name must not be empty";
+    Result<Positive<int>, string>  age  =
```
KaliCZ: Cna we also show the simplified way = x.AsPositive().ToTry("error")

readme.md
```diff
@@ -407,17 +406,102 @@ var missing =
...
+}
+```
+
+If both inputs are invalid, the aggregated error carries both messages:
+
+```csharp
+ParseUser("", -5);   // Error: "name must not be empty; age must be positive"
+```
+
+Fixed-arity overloads cover 2–8 inputs, plus an `IEnumerable` form for when the count is dynamic. The combiner overload (shown above) returns a built object directly; a tuple-returning form is also provided when there's no natural constructor to call.
```
KaliCZ: Don't use math terms like arity. Say that you can aggregate up to 8 task parameters or a an enumerable of them for validating lists for example. You can even show that as an example.

src/StrongTypes/Result/ResultFlattenExtensions.cs
```diff
@@ -0,0 +1,39 @@
...
+        where T : notnull
+        where TError : notnull
+        => nested.IsSuccess ? nested.InternalValue : nested.InternalError;
+
+    /// <summary>
+    /// Collapses a <see cref="Result{T}"/> of <see cref="Result{T}"/> into a single
+    /// <see cref="Result{T}"/>. Exists so the flattened value stays on the single-parameter
+    /// form instead of decaying to <c>Result&lt;T, Exception&gt;</c>.
+    /// </summary>
+    public static Result<T> Flatten<T>(this Result<Result<T>, Exception> nested)
```
KaliCZ, line 24: Shouldn't there be a TException and where TException : Exception?

And both the inner and outer should have the error type specified. So we might even have. a TInnerException and TOuterExcwption

src/StrongTypes/Result/ResultFlattenExtensions.cs
```diff
@@ -0,0 +1,39 @@
...
+        where T : notnull
+        => nested.IsSuccess ? nested.InternalValue : nested.InternalError;
+
+    /// <summary>
+    /// Concatenates an array-of-arrays error into a single flat array. Useful when
+    /// chaining <see cref="Result.Aggregate{T1, T2, TError}"/> calls — the outer
+    /// aggregation produces an error of <c>TError[][]</c> which this collapses back
+    /// to <c>TError[]</c>.
+    /// </summary>
+    public static Result<T, TError[]> FlattenErrors<T, TError>(this Result<T, TError[][]> r)
```
KaliCZ: I don't understand the usecase for this.
A result has 1 error. If I aggregate more of them, I'll have an array of errors. When will I have an array of arrays?

readme.md
```diff
@@ -407,17 +406,102 @@ var missing =
 
 `Maybe<T>` serializes via `System.Text.Json` as `{ "Value": x }` for `Some` and `{ "Value": null }` for `None` — no converter registration or custom `JsonSerializerOptions` needed. Deserialization also accepts `{}` for `None`, so callers can omit the property entirely.
 
-## Legacy types (to be replaced)
+### `Result<T, TError>`
```
KaliCZ, line 415: We should mention that try doesn't Serialize into json, because I saw no usecase for it.

It should be mentioned in this section and then also in the section that says that all types can Serialize into json.

PR #77, opened 2026-04-26
Interval strong types: inclusive-by-default bounds, JSON + EF Core + OpenAPI support

KaliCZ:
> [!WARNING]
> **Known performance regression — fix deferred to a follow-up PR ([#118](https://github.com/KaliCZ/StrongTypes/issues/118)).**
> This PR is the first to register an EF Core **materialization interceptor** (for interval read-validation and `AlwaysExclusive` bound restoration). Once any materialization interceptor is registered, EF routes *every* entity's materialization through the interceptor path — so enabling `UseStrongTypes()` regresses reads of **all** entities, even those with no interval, by roughly **6× time / 5× allocation in-process** (measured on a plain two-column entity; the allocation cost is provider-independent, the time cost is dwarfed by network latency on a real database). This is merged knowingly: the fix is scoped (only `AlwaysExclusive` genuinely needs the interceptor, so it should be registered conditionally) and tracked in **#118**.

## Summary

Adds four **interval** strong types — `FiniteInterval<T>`, `Interval<T>`, `IntervalFrom<T>`, `IntervalUntil<T>` — a pair of ordered endpoints with the invariant `Start <= End`, inclusive by default. The four `readonly struct` variants encode which endpoints are *bounded* in the type, so the compiler (not a runtime check) guarantees a bounded endpoint is never null. Endpoints are any `struct, IComparable<T>` (`int`, `DateTime`, `DateOnly`, `decimal`, …). Ships with JSON, EF Core, and OpenAPI support to match the rest of the library.

| Type | `Start` | `End` | For |
| --- | --- | --- | --- |
| `FiniteInterval<T>` | `T` | `T` | a fully-bounded range |
| `IntervalFrom<T>` | `T` | `T?` | "from X", no upper bound |
| `IntervalUntil<T>` | `T?` | `T` | "until Y", no lower bound |
| `Interval<T>` | `T?` | `T?` | either/both bounds optional |

## Core (`Kalicz.StrongTypes`)

- `TryCreate` / `Create` factories (reject `Start > End` and equal endpoints with an exclusive bound); bounds are inclusive by default, excludable per value via `startInclusive` / `endInclusive`.
- `Contains`, `Overlaps` / `GetOverlap` (allocation-free — the interval types are stack structs), `Deconstruct` for exhaustive switching over the four bound cases.
- **Implicit widening** to less-constrained variants (lossless); partial narrowing via the `As…` (nullable) / `To…` (throwing) convention.
- **Date bridging** between `DateTime` and `DateOnly`: cross-type `Contains`, `ToDateInterval()`, `Days()`, plus `DateOnly.ToTimeInterval()` and `DateTime.ToDateOnly()`.
- **JSON**: serialize as `{ "Start", "End" }`; inclusivity flags are written only when `false` and default to `true` on read. `IntervalJsonConverter<TInterval>` pins a bound's inclusivity per property (subclass with the two modes) instead of carrying it per value.

## EF Core (`Kalicz.StrongTypes.EfCore`)

- Two persistence shapes: **two indexable endpoint columns** (the `UseStrongTypes()` default) or a **single JSON column** (`HasIntervalJsonConversion`). Endpoint access in LINQ translates server-side (plain column, or `JSON_VALUE` / `->>`).
- **`IntervalBoundMode`** per bound — `AlwaysInclusive` (default, no flag column), `AlwaysExclusive` (fixed, no column, restored on read), `Stored` (own `bit` column, queryable).
- Integrity interceptor re-validates on read and enforces the bound mode on save (reflection-free compiled delegates; only runs for an interval property that actually changed).
- Inline endpoint column naming via `HasIntervalColumns(…, startName:, endName:)`, non-public (DDD) backing-property support, and a shadow discriminator so a `null` nullable interval stays distinct from a stored all-`NULL` unbounded interval.

## OpenAPI (`Kalicz.StrongTypes.OpenApi.Microsoft` / `.Swashbuckle`)

Interval schema renders as an object with `Start` / `End` and optional `StartInclusive` / `EndInclusive` booleans (the schema pins no default, since an omitted flag's inclusivity is the applied converter's choice), with only each variant's bounded endpoints in `required`.

## Testing

- Unit tests for the types, date bridging, JSON (per-value + pinned modes), and allocation micro-benchmarks.
- API integration tests round-trip every variant through the ASP.NET Core pipeline and EF Core against **SQL Server and PostgreSQL** (Testcontainers): the full column-mapping/bound-mode matrix, filtering/ordering, read-validation of corrupted rows across all four variants, and CRUD/PATCH parity with the scalar harness.
- OpenAPI document tests assert the schema across all four variants × three pipelines.

## Follow-ups

- **[#118](https://github.com/KaliCZ/StrongTypes/issues/118)** — the read-performance regression above (register the materialization interceptor conditionally).
- **[#116](https://github.com/KaliCZ/StrongTypes/issues/116)** — unify the scalar and interval CRUD/PATCH test harnesses to enforce feature parity by construction.


KaliCZ:
Rebased onto latest `main` and added a hardening commit. Highlights:

- **Error-path fix (matches #106):** the converter's nested endpoint reads lost the property path — a `null`/type-mismatch endpoint surfaced the validation error at the document root `$` instead of `$.value`. Now rethrows path-less so System.Text.Json reattaches the path, exactly as #106 did for the numeric converter. Error messages also no longer leak the `` ClosedInterval`1 `` arity name.
- **API integration tests:** four interval entities/controllers mapped via `HasIntervalJsonConversion`; wire→DB round-trip on SQL Server + PostgreSQL, null handling, update, and invalid payloads asserted at `$.value`.
- **OpenAPI:** the original PR shipped no schema support (intervals rendered opaque). Added a Microsoft schema transformer and a Swashbuckle filter rendering `{ Start, End }` with `required: [Start, End]` and per-variant endpoint nullability; verified on Microsoft 3.0, Microsoft 3.1, and Swashbuckle.
- **Expanded unit tests:** `Contains`/`ToString` per variant, JSON edge cases (missing/extra/reordered keys, non-object tokens, `DateOnly`/`DateTime` endpoints), and generator-branch coverage facts.
- **Docs:** README, SKILL.md + `references/intervals.md`, and the EfCore/OpenApi references and package readmes.

⚠️ **Correction to the description above:** EF Core `ComplexProperty` is **not** a supported persistence shape — EF can't bind the interval's private constructor as a complex type, and column-by-column materialization would bypass the `Start <= End` invariant. The single JSON column (`HasIntervalJsonConversion`, which round-trips through the validating converter) is the supported shape; docs were corrected accordingly.

Full build passes with `TreatWarningsAsErrors=true`; unit + API + OpenAPI suites green (API/OpenAPI run locally with `STRONGTYPES_SKIP_SQLSERVER=1` on an ARM64 box — CI exercises the real SQL Server path).

src/StrongTypes/Intervals/IntervalDateExtensions.cs
```diff
@@ -0,0 +1,88 @@
...
+    {
+        if (end is not { } e)
+        {
+            return null;
+        }
+        var day = DateOnly.FromDateTime(e);
+        return !endInclusive && e.TimeOfDay == TimeSpan.Zero && day != DateOnly.MinValue ? day.AddDays(-1) : day;
+    }
+
+    private static DateOnly? ToDate(DateTime? value) => value is { } v ? DateOnly.FromDateTime(v) : null;
```
KaliCZ: Don't we have a ToDateOnly extension? If not, we should have it - This is useful as a DateTime extension

src/StrongTypes/Intervals/IntervalDateExtensions.cs
```diff
@@ -0,0 +1,88 @@
...
+        Interval.Create(ToDate(interval.Start), CoveredEndDay(interval.End, interval.EndInclusive));
+
+    /// <summary>The number of calendar days the interval contains; an excluded endpoint day is not counted.</summary>
+    [Pure]
+    public static int Days(this FiniteInterval<DateOnly> interval) =>
+        interval.End.DayNumber - interval.Start.DayNumber + 1 - (interval.StartInclusive ? 0 : 1) - (interval.EndInclusive ? 0 : 1);
+
+    private static bool CoversDay(Interval<DateTime> interval, DateOnly day) => interval.Overlaps(DayWindow(day));
+
+    private static Interval<DateTime> DayWindow(DateOnly day) =>
```
KaliCZ: This is a useful extension - `ToTimeInterval` - again, start some real extensions for DateONly if we don't have any yet.

And this should be a Finite interval.

src/StrongTypes/Intervals/IntervalDateExtensions.cs
```diff
@@ -0,0 +1,88 @@
...
+    [Pure]
+    public static bool Contains(this IntervalFrom<DateTime> interval, DateOnly day) => CoversDay(interval, day);
+
+    /// <summary>Whether the interval covers any instant of the calendar day <paramref name="day"/>.</summary>
+    [Pure]
+    public static bool Contains(this IntervalUntil<DateTime> interval, DateOnly day) => CoversDay(interval, day);
+
+    /// <summary>Whether the interval covers any instant of the calendar day <paramref name="day"/>.</summary>
+    [Pure]
+    public static bool Contains(this Interval<DateTime> interval, DateOnly day) => CoversDay(interval, day);
```
KaliCZ, line 44: A lot of intervals get created just for us to be able to answer Contains. 
- Implicit Operator to Interval
- DayWindow
- Overlaps also creates an interval I believe. 

Would it be possible to simply implement the GetOverlap and Overlaps as methods that accept a Start and End? Then we don't need the implicit conversion and temporary wrappers. 

But I'm thinking about how much of an improvement that is, because the method will return a tuple and then we'll map the tuple result to an Interval if somebody asks for an Overlap. 

However the bool methods should probably benefit a lot anyway. 

Or am I overthinking this, because what happens on stack is super quick, so it doesn't matter?

src/StrongTypes.Api.IntegrationTests/Tests/ApiTests/Intervals/IntervalConventionMappingTests.cs
```diff
@@ -0,0 +1,210 @@
...
+/// endpoint columns by default — no explicit <c>HasIntervalColumns</c> — with a
+/// shadow discriminator on the nullable form, and an explicit
+/// <c>HasIntervalJsonConversion</c> opts into the single JSON column instead.
+/// Model-shape assertions only, against offline relational models: model building
+/// never opens a connection, so no connection string is configured and no
+/// containers are needed. Real-server behavior is covered by the Testcontainers
+/// suites; the InMemory provider is unusable even here because it maps any struct
+/// as a scalar, hiding the complex default.
+/// </summary>
+public class IntervalConventionMappingTests
```
KaliCZ, line 22: Thees tests are nice. But we're only verifying the contract. Not that the contract works. Why don't we try storing and reading some values? Perhaps ordering by some etc?

KaliCZ (commented):
Now that interval has the properties for Inclusive and Exclusive start/end, what will happen if I try to filter/sort by them into EF query? 

- I assume it should work when the column is mapped. 
- I assume it's gonna produce some unpredictable result with the JSON as it doesn't understand our defaulting.
- I assume it will crash if the column isn't mapped to EF core?

src/StrongTypes.Api.IntegrationTests/Tests/ApiTests/Intervals/IntervalEntityTests.cs
```diff
@@ -0,0 +1,337 @@
...
+/// The object-wire twin of the scalar <see cref="EntityTests{TSelf, TEntity, T, TNullable, TWire}"/>.
+/// An interval serializes as an object (<c>{ "start": …, "end": … }</c>) rather than a
+/// scalar, so it cannot reuse that harness's scalar <c>TWire</c> bodies and assertions —
+/// hence a parallel base. The two cover the <b>same functional surface and must not
+/// drift</b>: the create / get / update / PATCH scenarios here mirror it one-for-one, and
+/// a scenario added to either base must be added to the other. Only the invalid-payload
+/// cases differ by nature (endpoint order or a missing required endpoint here, a malformed
+/// scalar there).
+/// </remarks>
+public abstract class IntervalEntityTests<TEntity, TInterval>(TestWebApplicationFactory factory)
```
KaliCZ, line 27: For now, this is a base for interval specifically and when we have other objects that translate to a non-scalar object in JSON, then we'll split it into a generic base and interval buildup on top of it?

src/StrongTypes.Api.IntegrationTests/Tests/ConverterTests/Intervals/IntervalFilterTests.cs
```diff
@@ -0,0 +1,215 @@
...
+    public async Task RangeContainment_OnJsonColumn_TranslatesToSql(string provider)
+    {
+        SkipIfSqlServerUnavailable(provider);
+
+        var containing = await Seed(SqlDb.FiniteIntervalEntities, PgDb.FiniteIntervalEntities, FiniteInterval.Create(1, 10));
+        var disjoint = await Seed(SqlDb.FiniteIntervalEntities, PgDb.FiniteIntervalEntities, FiniteInterval.Create(20, 30));
+        var set = provider == "sql-server" ? SqlDb.FiniteIntervalEntities : PgDb.FiniteIntervalEntities;
+
+        var ids = await set
+            .Where(e => (e.Id == containing.Id || e.Id == disjoint.Id) && e.Value.Start <= 5 && 5 <= e.Value.End)
```
KaliCZ: Sounds like many of the tests could be simplified by a simple extension for tests to only filter entities with a specific ID (params, so that you can provide 3 for example) Then 
```
        var ids = await set
            .Where(e => e.Id == first.Id || e.Id == second.Id || e.Id == third.Id)
```
becomes
```
        var ids = await set
            .FilterById(first, second, third)
            .OtherLogic
```
Which is easy to read.

src/StrongTypes.Api.IntegrationTests/Tests/ConverterTests/Intervals/IntervalReadValidationTests.cs
```diff
@@ -0,0 +1,89 @@
...
+    {
+        var entityType = db.Model.FindEntityType(complexType.ContainingEntityType.ClrType)!;
+        var storeObject = StoreObjectIdentifier.Table(entityType.GetTableName()!, entityType.GetSchema());
+        return Quote(complexType.GetProperties().Single(p => p.Name == endpointName).GetColumnName(storeObject)!, provider);
+    }
+
+    private static string Quote(string identifier, string provider) =>
+        provider == "sql-server" ? $"[{identifier}]" : $"\"{identifier}\"";
+
+    private Task CorruptRow<TEntity>(DbContext db, string provider, Guid id, string setClause) where TEntity : class
```
KaliCZ, line 92: But we've onmly tested the Finite entity, right? Can we make sure we test all these specific logic types for all the types of intervals?

src/StrongTypes.EfCore/IntervalIntegrityInterceptor.cs
```diff
@@ -0,0 +1,320 @@
...
+            var corrected = rebuild.Invoke(
+                null, BindingFlags.DoNotWrapExceptions, binder: null, [value, startMode, endMode, displayName], culture: null);
+            if (!Equals(corrected, value))
+            {
+                propertyInfo.SetValue(instance, corrected);
+            }
+        };
+    }
+
+    private static Action<object>? BuildSaveChecker(Type entityClrType, IComplexProperty property)
```
KaliCZ: We shouldn't have to check the save - that's what our type checks. We only need to prevent reading invalid data.

src/StrongTypes.EfCore/IntervalIntegrityInterceptor.cs
```diff
@@ -0,0 +1,320 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.EntityFrameworkCore;
+using Microsoft.EntityFrameworkCore.Diagnostics;
+using Microsoft.EntityFrameworkCore.Metadata;
+
+namespace StrongTypes.EfCore;
+
+/// <summary>Guards column-mapped interval properties. On materialization it re-validates the interval invariant (a stored row violating it throws instead of producing an invalid interval) and applies the configured <see cref="IntervalBoundMode"/> to bounds that have no flag column. On save it rejects a value whose bound contradicts an <c>Always*</c> mode. (JSON-mapped intervals re-validate in their converter and carry their bounds in the payload.)</summary>
+internal sealed class IntervalIntegrityInterceptor : IMaterializationInterceptor, ISaveChangesInterceptor
```
KaliCZ, line 16: There's a lot of code in here - I hope we' wouldn't cause any performance issuess. This only runs when EF core actually marks a property as updated, right?

src/StrongTypes.EfCore/IntervalTypes.cs
```diff
@@ -0,0 +1,18 @@
+namespace StrongTypes.EfCore;
+
+internal static class IntervalTypes
+{
+    public static bool IsInterval(Type clrType)
```
KaliCZ: Is this only for EF core? I feel like I've seen this already

src/StrongTypes.EfCore/readme.md
```diff
@@ -37,6 +37,10 @@ services.AddDbContext<AppDbContext>(options => options
 2. Registers a method-call translator so `strongType.Unwrap()` inside a LINQ
    predicate rewrites as a plain column reference, retyped with the underlying
    type's mapping so downstream string/number operators compose cleanly.
+3. Registers integrity guards for column-mapped intervals: materializing a
+   stored row that violates `Start <= End` throws `InvalidOperationException`
+   instead of producing an invalid interval, and each property's
+   `IntervalBoundMode` is applied on read and enforced on save.
```
KaliCZ, line 43: Same question as in the other comment - if the user doesn't actualyl use intervals, then we're not gonna introduce some performance issues to him, right?  

I don't want to make Strongtypes have friction to be used by bigger projects

src/StrongTypes.OpenApi.IntegrationTests/Tests/OpenApiDocumentTestsBase.Intervals.cs
```diff
@@ -0,0 +1,125 @@
...
+        AssertIntegerEndpoint(Property(value, "End"), nullable: false);
+    }
+
+    [Fact]
+    public async Task Interval_BoundFlags_Render_As_Optional_Booleans_Defaulting_To_True()
+    {
+        var value = await IntervalValueSchema("/interval-entities/closed");
+        AssertBoundFlag(Property(value, "StartInclusive"));
+        AssertBoundFlag(Property(value, "EndInclusive"));
+    }
```
KaliCZ: How is this the only test for Inclusive? There's so many combinations possible. They should be tested.

src/StrongTypes.OpenApi.Microsoft/Intervals/IntervalSchemaTransformer.cs
```diff
@@ -0,0 +1,59 @@
...
+
+        SchemaPaint.ClearWrapperShape(schema);
+        schema.Type = JsonSchemaType.Object;
+        schema.Properties = new Dictionary<string, IOpenApiSchema>
+        {
+            ["Start"] = startSchema,
+            ["End"] = endSchema,
+            ["StartInclusive"] = BoundFlagSchema(),
+            ["EndInclusive"] = BoundFlagSchema(),
+        };
```
KaliCZ, line 38: Doesn't this mean that StartInclusive and EndInclusive will always be in the schema? I thought we don't want them in the schema by default. At least show them as nullable and optional

So supplying this to the JSON payload will actually work always through the API? We just said that the default JSON converter defaults to Inclusive.

src/StrongTypes.OpenApi.Swashbuckle/readme.md
```diff
@@ -42,6 +42,9 @@ app.UseSwaggerUI();
   `{ "type": "array", "minItems": 1, "items": <T schema> }`
 - `Maybe<T>` &rarr; object wrapper `{ "Value": <T schema> }` matching the
   converter's on-the-wire format.
+- `FiniteInterval<T>` / `Interval<T>` / `IntervalFrom<T>` / `IntervalUntil<T>` &rarr;
+  `{ "type": "object", "properties": { "Start": <T>, "End": <T> }, "required": ["Start", "End"] }`,
```
KaliCZ: And now I'm confused - I just saw the properties for Inclusivity in the schema. And you're saying that this is the output. So which is it?

src/StrongTypes.Tests/Intervals/IntervalJsonConverterModeTests.cs
```diff
@@ -0,0 +1,93 @@
...
+using System.Text.Json.Serialization;
+using Xunit;
+
+namespace StrongTypes.Tests;
+
+public class IntervalJsonConverterModeTests
+{
+    // The intended usage: a one-line subclass passing the two modes to the base.
+    private sealed class HalfOpenFiniteConverter()
+        : IntervalJsonConverter<FiniteInterval<int>>(IntervalBoundMode.AlwaysInclusive, IntervalBoundMode.AlwaysExclusive);
```
KaliCZ, line 11: Is this documented somewhere for the users to know about?

src/StrongTypes/Intervals/IntervalEndpoints.cs
```diff
@@ -0,0 +1,89 @@
+#nullable enable
+using System;
+
+namespace StrongTypes;
+
+// A null endpoint is unbounded; a bounded endpoint is inclusive or exclusive per its flag.
+internal static class IntervalEndpoints
```
KaliCZ: I'd call this helpers.

src/StrongTypes/Intervals/IntervalHelpers.cs
```diff
@@ -0,0 +1,89 @@
...
+    {
+        if (start is not { } s || end is not { } e)
+        {
+            return true;
+        }
+        var comparison = s.CompareTo(e);
+        return comparison < 0 || (comparison == 0 && startInclusive && endInclusive);
+    }
+
+    public static bool Contains<T>(T? start, T? end, bool startInclusive, bool endInclusive, T value) where T : struct, IComparable<T>
```
KaliCZ, line 19: So can't we use these methods also from the methods for DateOnly and DateTime? Then we can close Issue #117

src/StrongTypes/Intervals/IntervalJsonConverterFactory.cs
```diff
@@ -0,0 +1,238 @@
...
+using System.Linq.Expressions;
+using System.Reflection;
+using System.Text.Json;
+using System.Text.Json.Serialization;
+
+namespace StrongTypes;
+
+/// <summary><see cref="JsonConverterFactory"/> for the four interval types: <see cref="FiniteInterval{T}"/>, <see cref="Interval{T}"/>, <see cref="IntervalFrom{T}"/>, <see cref="IntervalUntil{T}"/>.</summary>
+/// <remarks>Each interval is read and written as a JSON object with <c>Start</c> and <c>End</c> properties. <c>StartInclusive</c> and <c>EndInclusive</c> are carried per value: written only when <c>false</c> and read as <c>true</c> when absent. To pin a bound's inclusivity instead, apply an <see cref="IntervalJsonConverter{TInterval}"/> to that property. Values that violate the wrapper's invariant throw <see cref="JsonException"/>. Property names honour the active <see cref="JsonNamingPolicy"/>.</remarks>
+public sealed class IntervalJsonConverterFactory : JsonConverterFactory
```
KaliCZ, line 13: Let's review this class for performance impact - if this performs badly, it could affect production APIs and that wouldn't be nice.

readme.md
```diff
@@ -26,6 +26,7 @@ StrongTypes adds small, focused types that make everyday code safer and more exp
 - [Helpful Types](#helpful-types)
   - [`NonEmptyString`](#nonemptystring)
   - [Numeric wrappers: `Positive<T>`, `NonNegative<T>`, `Negative<T>`, `NonPositive<T>`](#numeric-wrappers)
+  - [Interval types: `FiniteInterval<T>`, `Interval<T>`, `IntervalFrom<T>`, `IntervalUntil<T>`](#interval-types)
```
KaliCZ: I feel like I've said it multiple times. But let me say it again here. Interval types should sit right after the NonEmptyEnumerable and on the same level as NonEmptyENumerable. It should not be under the group Helpful Types. 

Because it needs to explain how it's serialized etc. 

So perhaps, we can rename Helpful types to Helpful Scalar types?

readme.md
```diff
@@ -104,6 +105,65 @@ All defaults (e.g. `default(Positive<T>)`) still satisfy their invariants (e.g.
...
+string label = interval switch
+{
+    (null, null)         => "unbounded",
+    (null, { } end)      => $"up to {end}",
+    ({ } start, null)    => $"from {start}",
+    ({ } start, { } end) => $"{start}..{end}",
+};
+```
+
+`Overlaps` / `GetOverlap` intersect intervals of any variant mix. Intervals that touch at a shared endpoint overlap in that single point when both touching bounds are inclusive — pass `endInclusive: false` for back-to-back time windows that must not overlap. `GetOverlap` returns `null` for disjoint intervals and keeps the receiver's bounded endpoints — a `FiniteInterval<T>` receiver yields `FiniteInterval<T>?`. `DateTime` and `DateOnly` intervals bridge across: a `DateTime` interval `Contains` a `DateOnly` when it covers any instant of that day (and the reverse goes by the moment's calendar day), `ToDateInterval()` converts to the days the interval covers, and `Days()` counts the days a `FiniteInterval<DateOnly>` contains.
```
KaliCZ: This needs to be way shorter.  You don't need to explain with text, showcase a method call. For example: people will understand what ToDateInterval does if you just show it.

readme.md
```diff
@@ -104,6 +105,65 @@ All defaults (e.g. `default(Positive<T>)`) still satisfy their invariants (e.g.
...
+
+```csharp
+FiniteInterval.Create(10, 20).GetOverlap(IntervalFrom.Create(15, null));   // [15, 20]
+stay.Contains(new DateOnly(2026, 7, 4));   // does the DateTime interval touch that day?
+```
+
+A more-constrained variant widens **implicitly** to a less-constrained one (`FiniteInterval<T>` → `IntervalFrom<T>` / `IntervalUntil<T>` / `Interval<T>`; `IntervalFrom<T>` and `IntervalUntil<T>` → `Interval<T>`) — the conversion is lossless (bound flags carry along) and never throws. That also lets you hold mixed variants in one collection by widening them to `Interval<T>` (e.g. `Interval<int>[] x = [finite, from, unbounded];`), stored inline as structs with no boxing. Narrowing back is partial, so it follows the `As…` convention and returns a nullable — `Interval<T>.AsFinite()` / `AsFrom()` / `AsUntil()` (and `AsFinite()` on the half-bounded variants) yield `null` when a required endpoint is unbounded, each with a throwing `To…` sibling (`ToFinite()`, …) for when an unbounded endpoint is a bug.
```
KaliCZ: Again, just showcase text is too long. Showcase 3 examples that work and showcase that you can't convert an Interval to FiniteInterval, because it doesn't compile. But you can use the `As` method to get it as nullable.

readme.md
```diff
@@ -104,6 +105,65 @@ All defaults (e.g. `default(Positive<T>)`) still satisfy their invariants (e.g.
...
+
+On the wire each interval is a JSON object `{ "Start": …, "End": … }` (both endpoint keys always present; an unbounded endpoint is `null`); `StartInclusive` / `EndInclusive` appear only when `false` and default to `true` when absent. Persist it with EF Core either as two scalar endpoint columns (the `UseStrongTypes()` default; each endpoint a plain, indexable column — bound inclusivity is a per-bound mapping choice via `IntervalBoundMode`: fixed with no extra column, or stored per value in its own column) or as a single JSON column, which round-trips through the validating converter, with endpoint access in LINQ translating to a server-side JSON path lookup. Both shapes re-validate the interval invariant on read, so a corrupted row throws instead of materializing an invalid interval.
```
KaliCZ: Json serialization now deserves a subheading. But again, too much text. Just show some code examples for registration. People will understand - the API is simple

readme.md
```diff
@@ -104,6 +105,65 @@ All defaults (e.g. `default(Positive<T>)`) still satisfy their invariants (e.g.
...
+
+```csharp
+// Default: two endpoint columns — rename them via the complex-property builder.
+var window = modelBuilder.Entity<Booking>().HasIntervalColumns(b => b.Window);
+window.Property(i => i.Start).HasColumnName("WindowStart");
+window.Property(i => i.End).HasColumnName("WindowEnd");
```
KaliCZ: Is this the right API? I thought we changed it so that you can do something like: 
```
var window = modelBuilder.Entity<Booking>().HasIntervalColumns(b => b.Window, startName: "WindowStart", endName: "WindowEnd");
```

And if we didn't do it yet - is there something stopping us from having a nice and simple API like that? Take a look at the registration methods overall and see if we could simplify.


On the other hand, also showcase one option specifying the inclusivity setting. And just link the docs for EF core package. But make sure the readme of EFCore package does actually specify all the possible combinations and how to achieve those.

readme.md
```diff
@@ -104,6 +105,65 @@ All defaults (e.g. `default(Positive<T>)`) still satisfy their invariants (e.g.
...
+
+// Or opt into a single JSON column instead.
+modelBuilder.Entity<Booking>().HasIntervalJsonConversion(b => b.Window);
+```
+
+See the [EF Core package readme](https://github.com/KaliCZ/StrongTypes/blob/main/src/StrongTypes.EfCore/readme.md).
```
KaliCZ: But make sure the readme of EFCore package does actually specify all the possible combinations and how to achieve those - like inclusivity etc.
```suggestion
See the [EF Core package readme](https://github.com/KaliCZ/StrongTypes/blob/main/src/StrongTypes.EfCore/readme.md) for more info about configuring interval storage.
```

readme.md
```diff
@@ -135,6 +195,10 @@ All strong types ship with `System.Text.Json` converters attached via `[JsonConv
 
 `Maybe<T>` has a special format of serialization, so Some serializes into `{ "Value": xxx }` and None into `{ "Value": null }`.
 
+The interval types serialize as an object `{ "Start": …, "End": … }` — on write, both endpoint keys are always present (an unbounded endpoint is `null`) and `StartInclusive` / `EndInclusive` appear only when `false`. On read, an absent key for an *optional* endpoint means `null` (so `Interval` accepts `{}`) and an absent bound flag means `true`; a payload that violates the invariant (`Start > End`, or equal endpoints with an exclusive bound) or omits/nulls a *required* endpoint surfaces as a `JsonException`.
+
+To pin a bound's inclusivity for a property instead of carrying it per value (mirroring the EF `IntervalBoundMode`), subclass `IntervalJsonConverter<TInterval>` with the two modes and apply it with `[JsonConverter]` — an `AlwaysInclusive` / `AlwaysExclusive` bound is never written and is forced on read: `public sealed class HalfOpenIntervalConverter() : IntervalJsonConverter<FiniteInterval<int>>(IntervalBoundMode.AlwaysInclusive, IntervalBoundMode.AlwaysExclusive);`.
```
KaliCZ: We should probably link to the file that actually does it. Is the test file understandable to a person who's never seen this?

readme.md
```diff
@@ -135,6 +195,10 @@ All strong types ship with `System.Text.Json` converters attached via `[JsonConv
 
 `Maybe<T>` has a special format of serialization, so Some serializes into `{ "Value": xxx }` and None into `{ "Value": null }`.
 
+The interval types serialize as an object `{ "Start": …, "End": … }` — on write, both endpoint keys are always present (an unbounded endpoint is `null`) and `StartInclusive` / `EndInclusive` appear only when `false`. On read, an absent key for an *optional* endpoint means `null` (so `Interval` accepts `{}`) and an absent bound flag means `true`; a payload that violates the invariant (`Start > End`, or equal endpoints with an exclusive bound) or omits/nulls a *required* endpoint surfaces as a `JsonException`.
```
KaliCZ: We can make this sentence simpler so that a 12 year old can understand it when learning how to code. As always, just show 2 example jsons - one without StartInclusive  - we can comment that Inclusive is true when skipped. And the second one will specify Inclusive. So you can show 2 code sections - one for reading and one for writing. Example speaks better than a long paragraph.

src/StrongTypes/Intervals/IntervalDateExtensions.cs
KaliCZ: Done in `0003938` — added public `DateTime.ToDateOnly()`, and the bridging helpers now use it.

src/StrongTypes/Intervals/IntervalDateExtensions.cs
KaliCZ: Done in `0003938` — added public `DateOnly.ToTimeInterval()` returning a `FiniteInterval<DateTime>` (finite, as you noted); `CoversDay` now goes through it.

src/StrongTypes.Api.IntegrationTests/Tests/ApiTests/Intervals/IntervalConventionMappingTests.cs
KaliCZ: The store/read/order coverage already exists in the Testcontainers matrix (`IntervalColumnMappingMatrixTestsBase`). In `0003938` I cross-referenced it from this file's XML doc and added offline flag-queryability tests here. This file stays model-shape-only by design (no container needed).

src/StrongTypes.EfCore/IntervalIntegrityInterceptor.cs
KaliCZ: Kept it — the benchmark shows it's cheaper than writing one column, so it's not the cost. In `2b7d941` I compiled away the reflective invoke (default insert overhead +21% -> +2% vs plain endpoint columns) and added a modified-skip so it only runs when the interval actually changed. Removing it would trade a loud error for silent data change (an exclusive value saved into an AlwaysInclusive column reads back inclusive).

src/StrongTypes.EfCore/IntervalIntegrityInterceptor.cs
KaliCZ: The reflective `Invoke` is gone (compiled to a delegate per entity type, cached). `FindAnnotation` was already per-type/cached, not per-row. And it now only runs for a complex property EF reports as modified, so an update that doesn't touch the interval pays nothing. `2b7d941`.

src/StrongTypes.EfCore/readme.md
KaliCZ: The save path now short-circuits the whole change-tracker scan when a model maps no bound-enforced interval, so a project not using intervals pays nothing there. One honest residual: the read materialization interceptor still does a per-row cached lookup for every entity (interceptors can't filter by type), but that's a ~ns dictionary hit with no allocation. `2b7d941`.

src/StrongTypes.Api.IntegrationTests/Tests/ApiTests/Intervals/IntervalEntityTests.cs
KaliCZ: Yes, exactly. The two CRUD bases are split only by wire shape — `EntityTests` (scalar) and `IntervalEntityTests` (JSON object) — and must stay at feature parity (documented in testing.md / CLAUDE.md / CONTRIBUTING). #116 tracks unifying them behind one shared base so a new non-scalar-JSON type builds on it and parity is enforced by construction rather than convention.

src/StrongTypes.Tests/Intervals/IntervalJsonConverterModeTests.cs
KaliCZ: Yes — the Skill covers it under "Pinning a bound's inclusivity" in `Skill/references/intervals.md` (the inheritable `IntervalJsonConverter<TInterval>` with a worked `HalfOpenIntervalConverter` example), and the main readme's JSON section points at `IntervalJsonConverter<TInterval>` too.

src/StrongTypes.EfCore/readme.md
KaliCZ: Correction — I resolved this too quickly. The save path does short-circuit cleanly, but the benchmark shows the read side is worse than I claimed: registering the materialization interceptor makes reads ~4x slower and allocate ~4x, for *every* entity in a UseStrongTypes context (not just intervals). That's exactly the friction you were worried about. Filed #118 with the analysis and fix options — reopening this until that's addressed.

src/StrongTypes/Intervals/IntervalDateExtensions.cs
KaliCZ: Confirmed with a micro-benchmark (`IntervalOperationBenchmarks`, commit `0de3e1c`): Overlaps 1.2ns, GetOverlap 2.2ns, Contains 0.2ns, and the bridging ContainsCalendarDay 4.7ns — all **0 B allocated**, since the interval types are structs. I also split the intersection so Overlaps returns a bool without building the result interval GetOverlap materializes. So the temporaries are stack-only; closing #117.

src/StrongTypes/Intervals/IntervalEndpoints.cs
KaliCZ: Renamed `IntervalEndpoints` -> `IntervalHelpers` in `0de3e1c`.

src/StrongTypes/Intervals/IntervalHelpers.cs
KaliCZ: Yes — the DateOnly/DateTime Contains already route through `IntervalHelpers.Contains`, and Overlaps now computes on raw endpoints without building a result interval, so the bridging reuses the same helpers with no temporaries. Benchmarked at 0 B (above). Closed #117. (`0de3e1c`)

src/StrongTypes.EfCore/IntervalTypes.cs
KaliCZ: You had seen it — the same 'is this an interval + what are its endpoint types' logic was in three places (this EfCore `IntervalTypes`, the JSON factory's SupportedDefinitions/StartEndTypes, and OpenApi's TryGetIntervalEndpoints). Consolidated into one core `StrongTypes.IntervalTypes` (IsInterval / TryGetEndpoints) and routed all three through it; deleted this EfCore copy. (`541c7d5`)

src/StrongTypes.OpenApi.Microsoft/Intervals/IntervalSchemaTransformer.cs
KaliCZ: They're **optional** (not in `required`) with `default: true`, and the default JSON converter genuinely reads them on input, so showing them is correct — a client can send `StartInclusive: false` and it takes effect. `nullable` would be wrong (null isn't accepted). So the schema is intentional; the actual defect was the readmes omitting them, now fixed in `9694ce3`.

src/StrongTypes.OpenApi.Swashbuckle/readme.md
KaliCZ: The schema is right (flags present, optional, default true); the readme was wrong to omit them. Fixed both the Swashbuckle and Microsoft readmes in `9694ce3` to show `StartInclusive`/`EndInclusive` and note they default to true.

src/StrongTypes/Intervals/IntervalJsonConverterFactory.cs
KaliCZ: Reviewed. It's already built for the hot path: converters are cached per type (`ConcurrentDictionary` + `GetOrAdd`), and `Inner<>` uses compiled expression delegates (getters, TryCreate, presence) built once per generic instantiation as `static readonly` — no reflection per Read/Write. The only per-operation allocation is the naming-policy `ConvertName` (a few small strings), inherent to honoring `JsonNamingPolicy`. Production-safe; happy to add a serialize/deserialize benchmark if you want a hard number.

readme.md
KaliCZ: Trimmed to a code block showing GetOverlap plus the bridging calls (Contains/ToDateInterval/Days/ToTimeInterval/ToDateOnly). (`latest`)

readme.md
KaliCZ: Reworked to three examples: implicit widening, holding mixed variants in one array, and `FiniteInterval<int> bad = mixed[0];` shown as *does NOT compile*, with `AsFinite()` (nullable) and `ToFinite()` (throws) as the narrowing path.

readme.md
KaliCZ: Done — added `startName`/`endName` to `HasIntervalColumns` (`7ac3499`) and the readme now uses the inline form plus an `AlwaysExclusive` inclusivity example, and links the EF Core readme (which documents every bound-mode combination).

readme.md
KaliCZ: The link is in place, and the EF Core readme's 'Bound inclusivity' section documents all the AlwaysInclusive / AlwaysExclusive / Stored combinations.

readme.md
KaliCZ: Added a '#### Persistence' subheading and cut the prose to a one-liner leading into the registration code block. (`latest`)

readme.md
KaliCZ: Replaced the paragraph with two example JSONs — one with the flags omitted (commented as inclusive-when-skipped) and one with `EndInclusive: false` — leaving just a short note on the invalid/required-endpoint cases.

readme.md
KaliCZ: Linked the source: the per-property pin now points at [`IntervalJsonConverter`](src/StrongTypes/Intervals/IntervalJsonConverter.cs), with the subclass shown as a code block rather than an inline sentence.

src/StrongTypes.Api.IntegrationTests/Tests/ApiTests/Intervals/IntervalConventionMappingTests.cs
KaliCZ: See the top-level thread; the round-trip/order coverage is in the Testcontainers matrix.

src/StrongTypes.Api.IntegrationTests/Tests/ConverterTests/Intervals/IntervalFilterTests.cs
KaliCZ: Added a `FilterById(params entities)` test extension and rewrote the id-filter chains to use it (IN / = ANY translation preserved). Suite green (1261 passed).

src/StrongTypes.Api.IntegrationTests/Tests/ConverterTests/Intervals/IntervalReadValidationTests.cs
KaliCZ: Extended read-validation to all four variants — a generic base + per-variant subclasses covering both the two-column interceptor re-validation and the JSON-converter re-validation (reversed endpoints + missing-required-endpoint per variant). Green.

src/StrongTypes.OpenApi.IntegrationTests/Tests/OpenApiDocumentTestsBase.Intervals.cs
KaliCZ: Added a theory over all four variants x three OpenAPI pipelines asserting StartInclusive/EndInclusive are optional booleans defaulting true, absent from `required`, with `required` equal to each variant's bounded endpoints. 324 passing.

readme.md
KaliCZ: Done in `731ad08` — 'Interval types' is now a top-level section between NonEmptyEnumerable and Parsing helpers, and 'Helpful Types' is renamed 'Helpful Scalar types' (TOC updated to match).

src/StrongTypes/Intervals/IntervalDateExtensions.cs
```diff
@@ -0,0 +1,90 @@
...
+    /// <summary>The calendar day of <paramref name="moment"/>, dropping its time of day.</summary>
+    [Pure]
+    public static DateOnly ToDateOnly(this DateTime moment) => DateOnly.FromDateTime(moment);
+
+    /// <summary>The instants of the calendar day <paramref name="day"/> as the half-open interval [midnight, next midnight); the last representable day is capped at <see cref="DateTime.MaxValue"/>.</summary>
+    [Pure]
+    public static FiniteInterval<DateTime> ToTimeInterval(this DateOnly day) =>
+        day == DateOnly.MaxValue
+            ? FiniteInterval.Create(day.ToDateTime(TimeOnly.MinValue), DateTime.MaxValue)
+            : FiniteInterval.Create(day.ToDateTime(TimeOnly.MinValue), day.AddDays(1).ToDateTime(TimeOnly.MinValue), endInclusive: false);
```
KaliCZ: Let's move these three at the top before the Contains group starts

KaliCZ: Done in `c7a0765` — `Days`, `ToDateOnly`, and `ToTimeInterval` now sit at the top, above the `Contains` group.

PR #78, opened 2026-04-26
Add OpenApi support

KaliCZ:
## Summary

Adds two parallel NuGet packages that wire StrongTypes into the two ASP.NET Core OpenAPI pipelines, plus a pair of analyzers (ST0002/ST0003) that prompt a project to install the matching adapter. Both packages share a small internal painter/inliner library, mirror each other in surface, and produce the same wire-shape contract — including caller-supplied data annotations layered on top of the wrapper paint.

## What's in the box

| Package | For apps using | Extension point |
| --- | --- | --- |
| `Kalicz.StrongTypes.OpenApi.Microsoft` | `Microsoft.AspNetCore.OpenApi` (`AddOpenApi()`) | `IOpenApiSchemaTransformer` + `IOpenApiDocumentTransformer` |
| `Kalicz.StrongTypes.OpenApi.Swashbuckle` | `Swashbuckle.AspNetCore` (`AddSwaggerGen()`) | `ISchemaFilter` + `IDocumentFilter` |

The internal `OpenApi.Core` project is `IsPackable=false`; its DLL is embedded into each adapter package via `BuildOutputInPackage` so consumers don't see a stray `Kalicz.StrongTypes.OpenApi.Core` id on NuGet.

Both adapters expose `options.AddStrongTypes()` and produce the same wire JSON for every wrapper:

- `NonEmptyString` → `{ type: "string", minLength: 1 }`
- `Positive<T>` / `NonNegative<T>` / `Negative<T>` / `NonPositive<T>` → underlying primitive with the appropriate min/max bound
- `NonEmptyEnumerable<T>` / `INonEmptyEnumerable<T>` → `{ type: "array", minItems: 1, items: <T schema> }`
- `Maybe<T>` → object wrapper `{ Value: <T schema> }` — inlined at each use site by `StrongTypeInliner`, no named component leaks into `components.schemas`

## Surgical paint

The painter clears only the CLR-wrapper-derived shape (`Properties`, `Required`, `AllOf`/`OneOf`/`AnyOf`, `AdditionalProperties`, `Items`) before writing the wire primitive over the top. Caller-set fields (`Pattern`, `MinLength`, `MaxLength`, `Minimum`/`Maximum`/`ExclusiveMinimum`/`ExclusiveMaximum`, `Description`, `Example`, `Default`, `Enum`, custom extensions) are left alone. Where the wrapper has a wire-shape floor (`minLength: 1` for `NonEmptyString`, `exclusiveMinimum: 0` for `Positive<T>`, etc.), it is applied tighter-wins — only filling in or strengthening the caller's value, never relaxing or wiping it.

## Caller annotations on property positions

Both pipelines normally drop every `ValidationAttribute` on a property whose CLR type carries a custom `JsonConverter` — which every wrapper does — collapsing the property to a bare `$ref` to the wrapper component. A property-level pass on each pipeline (`PropertyAnnotationSchemaTransformer` for Microsoft, `PropertyAnnotationSchemaFilter` for Swashbuckle) re-applies them: **every annotation the underlying pipeline natively supports on a primitive-typed property also works on a wrapper-typed one.** The two passes wrap the `$ref` position in `{ allOf: [$ref], <bounds> }` so the caller bounds compose with the wrapper's floor under the same tighter-wins rules; `StrongTypeInliner` then flattens the `allOf+ref` into a single inline schema and removes the wrapper component from `components.schemas` afterwards.

The two adapters intentionally support different attribute sets — each mirrors its own pipeline's native behaviour, so the wrapper-typed surface stays consistent with the primitive-typed surface on that pipeline. `Microsoft.AspNetCore.OpenApi` natively maps more attributes (`[Length]`, `[Base64String]`, `[Description]`, `[Range(MinimumIsExclusive = …)]`, …) than Swashbuckle's `DataAnnotationsSchemaFilter` does; if you need the broader set, the Microsoft adapter is the one to install.

Worked examples on the wire (Microsoft):

| Property | Annotations | Result |
| --- | --- | --- |
| `NonEmptyString Username` | `[StringLength(50, MinimumLength=3)]` + `[RegularExpression(...)]` | `minLength:3` (caller > floor 1), `maxLength:50`, `pattern` |
| `NonEmptyString Email` | `[StringLength(254)]` + `[RegularExpression(...)]` | `minLength:1` (floor wins, caller didn't set), `maxLength:254`, `pattern` |
| `Positive<int> Age` | `[Range(18, 120)]` | `minimum:18`, `maximum:120` |
| `Positive<int> RangeAcrossFloor` | `[Range(-5, 5)]` | `exclusiveMinimum:0` (floor wins, beats caller's -5), `maximum:5` |
| `Positive<int> ExclusiveAtFloor` | `[Range(0, 5, MinimumIsExclusive = true)]` | `exclusiveMinimum:0` (annotation matches floor; deduplicated), `maximum:5` |
| `Positive<int> InclusiveJustAboveFloor` | `[Range(1, 5)]` | `minimum:1` inclusive (numerically tighter than `> 0`; exclusive form drops out), `maximum:5` |
| `Positive<int> RangeBelowFloor` | `[Range(-10, -5)]` | `exclusiveMinimum:0` (floor wins) + `maximum:-5` (caller verbatim) — unsatisfiable, but emitted as-is |
| `NonEmptyEnumerable<NonEmptyString> Tags` | `[MaxLength(10)]` | `minItems:1`, `maxItems:10` |
| `NonEmptyString ContactEmail` | `[EmailAddress]` | `minLength:1`, `format:"email"` |
| `NonEmptyString Website` | `[Url]` | `minLength:1`, `format:"uri"` |
| `NonEmptyString Tagline` | `[Description("…")]` | `minLength:1`, `description:"…"` |

## Required-array — parity with primitives, not normalisation

Earlier iterations rewrote each parent schema's `required` set from C# nullability so both pipelines agreed on a "non-nullable C# → required" semantic. That overstepped — it touched non-StrongTypes properties and ignored the C# `required` keyword and `[Required]` attribute that the underlying pipelines actually read. The override is gone; what's left is a parity contract: **for every input axis the pipeline cares about (plain non-nullable, plain nullable, `[Required]`, the C# `required` keyword), a strong-type property's membership in the `required` array matches the equivalent primitive property's membership.** Whatever the underlying pipeline emits is what the wrapper-typed surface mirrors.

## Document inliner

`StrongTypeInliner` (a single shared `Core` pass, run as `IDocumentFilter` on Swashbuckle and `IOpenApiDocumentTransformer` on Microsoft) walks the entire document after every other pass and:

1. Detects each wrapper component (`NonEmptyString`, `PositiveOf*`, `NonEmptyEnumerableOf*`, `MaybeOf*`, plus Swashbuckle's suffix-style equivalents like `Int32Positive`, `NonEmptyStringMaybe`).
2. Replaces every `$ref` to it (in property positions, array `items`, dictionary `additionalProperties`, response bodies, parameters) with the wrapper's wire body — merging caller-supplied annotations attached at the use site under tighter-wins.
3. Drops the wrapper component from `components.schemas` afterwards.

A new integration test (`Inlineable_Wrapper_Components_Are_Removed_From_Components_Schemas`) scans the generated document and fails on any leftover wrapper component, locking the cleanup contract in.

## Analyzers

Two new analyzers in `Kalicz.StrongTypes.Analyzers` (parallel to the existing ST0001 EFCore one):

- **ST0002** — fires when a project references `Microsoft.AspNetCore.OpenApi` and exposes a public StrongTypes-typed property, but doesn't reference `Kalicz.StrongTypes.OpenApi.Microsoft`.
- **ST0003** — same shape for `Swashbuckle.AspNetCore` ↔ `Kalicz.StrongTypes.OpenApi.Swashbuckle`.

Both ship with a code fix that appends the matching `<PackageReference>` to the project's csproj — same disk-side-effect pattern as the EFCore code fix.

## OpenAPI 3.0 and 3.1

The Microsoft pipeline test app pinned 3.0; a parallel `MicrosoftTestApi31Factory` re-runs the entire shape contract under 3.1 so the two encodings of exclusive bounds (`minimum:0, exclusiveMinimum:true` boolean-paired vs `exclusiveMinimum:0` numeric) and nullable-ref unions (`oneOf` + `nullable:true` vs `anyOf` + `type:"null"`) are both verified.

## Test layout

`StrongTypes.OpenApi.IntegrationTests` hosts both pipelines' test apps directly and runs a single shared assertion suite three times — once per concrete subclass (`MicrosoftOpenApiDocumentTests`, `MicrosoftOpenApi31DocumentTests`, `SwashbuckleOpenApiDocumentTests`) — so any divergence between Microsoft 3.0, Microsoft 3.1, or Swashbuckle output shows up as a per-generator failure. The shared suite is split into feature-scoped partials (Strings, Numerics, Collections, Composition, Annotations, Components) so each concern opens to roughly a screen of related tests, with nullable variants sitting next to their non-nullable siblings. A single per-feature broken-pipeline flag (`IsEmailStringFormatBroken`) lets tests assert the keyword is *absent* on the pipeline that doesn't natively honour it, rather than skipping — so the suite documents the framework's actual behaviour and catches it if the framework starts honouring the annotation. The DTOs and controllers live in a shared `StrongTypes.OpenApi.TestApi.Shared` library and are mounted into both test apps via `AddApplicationPart`, guaranteeing the on-the-wire shapes the tests assert against are byte-for-byte identical between pipelines.

`StrongTypes.Api` no longer carries any OpenAPI wiring — its `Api.IntegrationTests` continues to focus on EF Core round-tripping against SQL Server + PostgreSQL containers.

## Docs

- Per-package readmes for both adapters with install/register snippets, the wire-shape table, and a Data annotations section.
- Top-level `readme.md` gets a `## Packages` index linking to every per-package readme; pack-time `_RewriteReleaseUrls` rewrites those relative `src/...` links to `blob/v<version>/...` URLs pinned to the release tag, so nuget.org-rendered readmes always point at the asset that shipped with that version.
- Top-level readme + skill `references/openapi.md` recommend Swashbuckle when the choice is open, since `Microsoft.AspNetCore.OpenApi` has rough edges the framework gives no public hook to fix (`[EmailAddress]` doesn't propagate as `format: email`, dictionary keys with strong types aren't always honoured, …).
- Every per-package readme carries the full set of five download badges so each package's nuget.org page advertises the suite.
- `docs/diagrams/package-layout.svg` updated for the renamed packages and a fleshed-out FsCheck card.
- Skill `SKILL.md` lists both adapter packages; `references/openapi.md` covers wiring for both pipelines plus the wrapper-property data-annotation contract.

## Test plan

- [x] 192 OpenAPI document tests across all 3 pipelines pass (Microsoft 3.0, Microsoft 3.1, Swashbuckle) — no containers
- [x] EF Core integration tests pass on `Api.IntegrationTests`
- [x] 39 analyzer tests pass (15 EFCore + 15 OpenAPI analyzer + 5 EFCore code-fix + 4 OpenAPI code-fix)
- [x] Full unit-test suite passes
- [x] Both adapter `dotnet pack` outputs embed `StrongTypes.OpenApi.Core.dll` without surfacing it as a NuGet dependency
- [x] Packed readme links resolve to `blob/v<version>/...` URLs (verified with `dotnet build -c Release -p:Version=9.9.9`)


src/StrongTypes.OpenApi.IntegrationTests/Tests/OpenApiDocumentTestsBase.Annotations.cs
```diff
@@ -0,0 +1,292 @@
...
+        var doc = await GetDocumentAsync();
+        var body = FollowRef(doc, RequestSchema(doc, "/annotated-numbers"));
+        var exclusive = Property(body, "exclusiveLowerAge");
+
+        // [Range(2, 10, MinimumIsExclusive = true)] degrades to plain
+        // [Range(2, 10)] when MinimumIsExclusive is dropped — minimum
+        // becomes inclusive 2.
+        Assert.Equal(10m, CollectMinUpperBound(doc, exclusive, Version));
+        if (IsExclusiveRangeBroken)
+            Assert.Equal(2m, CollectMaxLowerBound(doc, exclusive, Version));
```
KaliCZ: Follow-up: rather than baking `IsExclusiveRangeBroken` into the suite for Swashbuckle, we should fix the pipeline so the exclusive flag survives. We re-apply `[Range]` ourselves in the wrapper-annotation pass, so the flag dropping is almost certainly on us — `MinimumIsExclusive` (and `MaximumIsExclusive`) is being read but not written through. Once that lands, this branch and the flag both go away.

KaliCZ: Update / correction: verified end-to-end with a fresh ASP.NET project — vanilla Swashbuckle does emit `minimum: 2, exclusiveMinimum: true` for `[Range(2, 10, MinimumIsExclusive = true)] int`, and so does our pipeline. The earlier broken-flag branch was a side-effect of using `Range(1, 10)`: the wrapper paint emits `minimum: 0, exclusiveMinimum: true` and the property-level Range emits `minimum: 1, exclusiveMinimum: true`, which the inliner could collapse to the wrapper layer's bound (0) instead of the annotation's (1). Shifting the example to `Range(2, 10)` makes the annotation strictly tighter, so the bound survives unambiguously. Flag and broken-branch are gone in 6f6ed6c.

PR #91, opened 2026-05-04
Strong-type schema painting and annotation merging on non-body slots (closes #87)

### LINKED ISSUE #87: Microsoft.AspNetCore.OpenApi: rewrite non-body parameter schemas for strong types
Follow-up to #84.

`Kalicz.StrongTypes.OpenApi.Microsoft` currently only rewrites schemas reachable through the JSON-body code path (where `IOpenApiSchemaTransformer` runs against `JsonTypeInfo`). Schemas attached to non-body parameters — `[FromQuery]`, `[FromRoute]`, `[FromHeader]`, `[FromForm]` — go through a different schema-generation path and the strong-type transformers never see them. They come out as the underlying primitive (and sometimes not even that — `Positive<int>` from `[FromQuery]` renders as `{"type":"string"}` because the binder hint is the `IParsable.TryParse` string overload).

`StrongTypes.OpenApi.IntegrationTests/Tests/OpenApiDocumentTestsBase.NonBodyBinding.cs` documents the gap with `IsNonBodyParameterStrongTypeSchemaBroken = true` / `IsFormPropertyStrongTypeSchemaBroken = true` on the Microsoft subclass — fixing this means flipping those flags off and turning the broken-path branches into a no-op.

## What's needed

An `IOpenApiOperationTransformer` (or document transformer that walks operations) that:

1. Iterates `paths.<path>.<method>.parameters[]` and replaces each parameter's `schema` with the strong-type schema when the parameter's .NET type is a strong-type wrapper.
2. For `requestBody` of `multipart/form-data` / `application/x-www-form-urlencoded`, walks `properties` and rewrites each property's schema by the same rule.
3. Reuses the existing `SchemaPaint` / wrapper-painter helpers in `StrongTypes.OpenApi.Core` so the wire shape is identical to the JSON-body path.

The metadata needed to map a parameter back to its .NET type lives on `OpenApiOperationTransformerContext.Description.ParameterDescriptions`; each `ApiParameterDescription` exposes the `Type`.

## Done when

- `IsNonBodyParameterStrongTypeSchemaBroken` and `IsFormPropertyStrongTypeSchemaBroken` are removed from the Microsoft subclasses (`MicrosoftOpenApiDocumentTests`, `MicrosoftOpenApi31DocumentTests`).
- The shared `OpenApiDocumentTestsBase.NonBodyBinding` tests pass against the Microsoft pipeline on the strong-type-aware branch.

KaliCZ:
## Summary

Started as the issue #87 fix (Microsoft non-body schema painting). Grew during review to land symmetrical, complete annotation handling on both pipelines and a few related cleanups.

### What's in
- **Microsoft non-body schema painting (the original #87 ask).** New `NonBodyStrongTypeOperationTransformer` walks `ApiDescription.ParameterDescriptions`, resolves each slot's CLR type via `ModelMetadata.ModelType` (the only source `IParsable<T>` doesn't mask), and paints the wrapper's wire shape directly onto `[FromQuery]` / `[FromRoute]` / `[FromHeader]` parameter schemas and `[FromForm]` property schemas.
- **Caller annotation merging on non-body slots, both pipelines.** Both transformers/filters now run `WrapperAnnotationApplier` after the wire-shape paint so a `[FromQuery, StringLength(50)] NonEmptyString name` reaches the wire as `{type:string, minLength:1, maxLength:50}` instead of dropping the caller's `maxLength`. The applier moved from the Microsoft adapter to `StrongTypes.OpenApi.Core` so both pipelines share the same logic. Microsoft mutates its slot in place; Swashbuckle wraps the `$ref` in `allOf:[<ref>] + annotations` and lets the existing inliner collapse it at the end, mirroring the body-side trick `PropertyAnnotationSchemaFilter` already uses.
- **Email type wired into the annotation appliers.** Email was missing from both `WrapperAnnotationApplier` (Microsoft) and `PropertyAnnotationSchemaFilter` (Swashbuckle), so `[StringLength(100)] Email` silently dropped the caller's `maxLength: 100` on both pipelines for body and non-body slots. Email is string-shaped, so the same `ApplyStringAnnotations` branch handles it. Body-side annotation test added.
- **Primitive baselines on every annotated wrapper test.** Each `[FromQuery]` / `[FromForm]` wrapper test now has a `string` / `int` sibling carrying the same annotation. Sets the floor: a future framework regression would fail the baseline first, so a wrapper failure can be correctly attributed instead of guessed at.
- **Cleanup.** `IsNonJsonBodyStrongTypeSchemaBroken` and `IsNonBodyAnnotationMergingBroken` test flags both deleted (no pipeline overrides them anymore). `BindingSchemaAsserts` helpers renamed from `Assert{NonBody,Route}*` to `Assert*Schema` (the wire shape doesn't depend on body/non-body — the old names implied otherwise). Misleading XML comments rewritten.

### Body conflict safety

The non-body painters are insulated from JSON body schemas three ways: (1) source filter dispatches only on `BindingSource.Query | Path | Header | Form` — never `Body`; (2) the form path's content-type whitelist is `multipart/form-data` and `application/x-www-form-urlencoded` only — `application/json` is never iterated; (3) structurally, `operation.Parameters[]` only contains non-body slots in OpenAPI/ASP.NET (body content lives under `operation.RequestBody`), so the parameter-rewrite path can't reach a JSON body schema even if the source filter were bypassed. The unchanged 200+ JSON-body tests in the Annotations / Strings / Numerics / Components partials act as the regression lock.

### Precision

Type matching is exact (`==` on `Type` and on `GetGenericTypeDefinition()`); user DTOs never match. Form properties are checked individually — only strong-type slots in a mixed form are rewritten. Form-property keys are looked up by exact name (Microsoft's PascalCase) or single camelCase fallback (Swashbuckle's convention) — no speculative dot-splitting or case-insensitive scans that could bind to the wrong slot.

## Test plan
- [x] Full repo suite: 2,185 tests green across `StrongTypes.Tests` (1,038), `StrongTypes.Analyzers.Tests` (39), `StrongTypes.OpenApi.IntegrationTests` (279), and `StrongTypes.Api.IntegrationTests` (829).
- [x] +27 new tests over the original baseline: non-body wrapper schema painting, primitive baselines, wrapper annotation merging, Email annotation merging, all running on both Microsoft and Swashbuckle pipelines.

### DIFF
--- a/src/StrongTypes.OpenApi.Core/Inlining/StrongTypeInliner.cs
+++ b/src/StrongTypes.OpenApi.Core/Inlining/StrongTypeInliner.cs
@@ -139,6 +139,8 @@ public static class StrongTypeInliner
 
         if (schema.Not is { } not)
             schema.Not = RewriteSlot(not, ctx);
+
+        StrongTypeInlineMarker.Remove(schema);
     }
 
     private static void RewriteVariantList(IList<IOpenApiSchema>? variants, RewriteContext ctx)
--- a/src/StrongTypes.OpenApi.Core/StrongTypeSchemaTypes.cs
+++ b/src/StrongTypes.OpenApi.Core/StrongTypeSchemaTypes.cs
@@ -7,7 +7,7 @@ public static class StrongTypeSchemaTypes
 
     public static bool IsInlineable(Type? clrType)
     {
-        return ResolveAnnotationSurrogate(clrType) is not null
+        return ResolveWireType(clrType) is not null
             || TryGetMaybeValue(clrType, out _);
     }
 
@@ -59,7 +59,7 @@ public static class StrongTypeSchemaTypes
         return true;
     }
 
-    public static Type? ResolveAnnotationSurrogate(Type? clrType)
+    public static Type? ResolveWireType(Type? clrType)
     {
         if (IsNonEmptyString(clrType) || IsEmail(clrType)) return typeof(string);
         if (IsDigit(clrType)) return typeof(int);
--- a/src/StrongTypes.OpenApi.Core/StrongTypeSchemaTypes.cs
+++ b/src/StrongTypes.OpenApi.Core/StrongTypeSchemaTypes.cs
@@ -7,15 +7,8 @@ public static class StrongTypeSchemaTypes
 
     public static bool IsInlineable(Type? clrType)
     {
-        var unwrapped = UnwrapNullable(clrType);
-        if (unwrapped is null) return false;
-
-        return IsNonEmptyString(unwrapped)
-            || IsEmail(unwrapped)
-            || IsDigit(unwrapped)
-            || TryGetNumeric(unwrapped, out _, out _)
-            || TryGetNonEmptyEnumerableElement(unwrapped, out _)
-            || TryGetMaybeValue(unwrapped, out _);
+        return ResolveAnnotationSurrogate(clrType) is not null
+            || TryGetMaybeValue(clrType, out _);
     }
 
     public static bool IsNonEmptyString(Type? clrType)
@@ -64,5 +57,14 @@ public static class StrongTypeSchemaTypes
 
         valueType = unwrapped.GetGenericArguments()[0];
         return true;
+    }
+
+    public static Type? ResolveAnnotationSurrogate(Type? clrType)
+    {
+        if (IsNonEmptyString(clrType) || IsEmail(clrType)) return typeof(string);
+        if (IsDigit(clrType)) return typeof(int);
+        if (TryGetNumeric(clrType, out var valueType, out _)) return valueType;
+        if (TryGetNonEmptyEnumerableElement(clrType, out var elementType)) return typeof(IEnumerable<>).MakeGenericType(elementType);
+        return null;
     }
 }
--- a/src/StrongTypes.OpenApi.Core/StrongTypeSchemaTypes.cs
+++ b/src/StrongTypes.OpenApi.Core/StrongTypeSchemaTypes.cs
@@ -1,0 +1,19 @@
+namespace StrongTypes.OpenApi.Core;
+
+public static class StrongTypeSchemaTypes
+{
+    public static bool IsInlineable(Type? clrType)
+    {
+        if (clrType is null) return false;
+
+        var unwrapped = Nullable.GetUnderlyingType(clrType) ?? clrType;
+        if (unwrapped == typeof(NonEmptyString) || unwrapped == typeof(Email) || unwrapped == typeof(Digit)) return true;
+        if (!unwrapped.IsGenericType) return false;
+
+        var definition = unwrapped.GetGenericTypeDefinition();
+        return NumericWrapperKinds.TryGetBound(definition, out _)
+            || definition == typeof(NonEmptyEnumerable<>)
+            || definition == typeof(INonEmptyEnumerable<>)
+            || definition == typeof(Maybe<>);
+    }
+}
--- a/src/StrongTypes.OpenApi.Core/StrongTypeSchemaTypes.cs
+++ b/src/StrongTypes.OpenApi.Core/StrongTypeSchemaTypes.cs
@@ -2,18 +2,67 @@ namespace StrongTypes.OpenApi.Core;
 
 public static class StrongTypeSchemaTypes
 {
+    public static Type? UnwrapNullable(Type? clrType)
+        => clrType is null ? null : Nullable.GetUnderlyingType(clrType) ?? clrType;
+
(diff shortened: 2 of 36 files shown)


KaliCZ:
Pushed `b9a41c4` addressing both review points.

**Body-conflict safety (your most recent question)** — there are two layers of filtering and one structural reason that together ensure the transformer never reaches a JSON body:
1. Source filter: dispatches only on `BindingSource.Query | Path | Header | Form`. `BindingSource.Body` is never matched, and ASP.NET classifies sources mutually-exclusively.
2. Content-type filter on the form path: `RewriteFormProperty` only iterates `multipart/form-data` and `application/x-www-form-urlencoded`. JSON bodies live at `application/json`.
3. Structural: in OpenAPI, `operation.Parameters[]` only contains non-body slots — body content lives in a separate `operation.RequestBody`. The parameter-rewrite path *cannot* reach a JSON body schema.

The full unchanged JSON-body / Annotations / Components test suite (236 tests across both pipelines) acts as the regression lock — if my transformer accidentally rewrote a body schema, those would have started failing.

**Annotation merging** — non-body slots now run `WrapperAnnotationApplier` after the wrapper paint, the same helper the JSON-body pipeline uses. Microsoft transformer mutates `parameter.Schema` in place rather than replacing it, so any keywords the pipeline had already written (description, default, the caller-applied `[StringLength]` Microsoft attaches to the IParsable string overload, etc.) survive. Four new tests pin this:
- `FromQuery_NonEmptyString_With_StringLength_Carries_Both_Bounds_When_Merged`
- `FromQuery_PositiveInt_With_Range_Carries_Both_Bounds_When_Merged`
- `FromForm_NonEmptyString_With_StringLength_Carries_Both_Bounds_When_Merged`
- `FromForm_PositiveInt_With_Range_Carries_Both_Bounds_When_Merged`

Swashbuckle's pipeline doesn't merge non-body annotations; new `IsNonBodyAnnotationMergingBroken` flag (overridden true in `SwashbuckleOpenApiDocumentTests`) makes the broken path assert the wrapper-only shape, so the suite catches it the day Swashbuckle starts honouring them.

**Dead-flag cleanup** — `IsNonJsonBodyStrongTypeSchemaBroken` removed entirely along with the broken-path branches in `BindingSchemaAsserts`.

Full repo suite: 2,170 tests green (was 2,158; +12 annotation tests, 6 per pipeline).

KaliCZ:
Pushed `2a1a18f`. Recap of the four review points:

**1. The misleading "leftover history" XML comment** — gone. The new `BindingSchemaAsserts` doc just describes what the helpers do today (centralised wire-shape assertions + form navigation), without referring to deleted branches.

**2. Helper names** — you were right that `AssertNonBody*` implied a body-vs-non-body shape difference that doesn't exist. The wire shape is identical regardless of where the schema sits. Renamed to:
- `AssertNonEmptyStringSchema` / `AssertPositiveIntSchema` / `AssertEmailSchema`
- `AssertFormPropertyNonEmptyStringSchema` / `AssertFormPropertyPositiveIntSchema` / `AssertFormPropertyEmailSchema`

The form-property variants stay because they do real navigation work (broken-allOf vs proper-properties); the parameter variants stay because they centralise the literal shape.

**3. `IsNonBodyAnnotationMergingBroken` — is it framework or ours?** Investigated with primitive baselines, as you suggested. Both pipelines emit `{maxLength:50, minLength:0, type:string}` for `[FromQuery, StringLength(50)] string plainName`. So the framework supports it on primitives — Swashbuckle dropping it on `NonEmptyString` is **our** adapter's gap, not Swashbuckle's. The flag's doc now says exactly that and points to the baseline tests as proof. A follow-up task is queued (chip in this session) to implement the missing pass in the Swashbuckle adapter and flip the flag off.

**4. `ResolveFormPropertyKey` dot-splitting** — also gone. ASP.NET / Microsoft.AspNetCore.OpenApi key form properties by the C# property name (PascalCase), matching `ApiParameterDescription.Name` exactly. Per CLAUDE.md "don't add fallbacks for scenarios that can't happen", and the speculative fallbacks risked binding to the wrong slot if a user DTO ever had near-collisions.

**Bonus — split annotated form bodies.** The new annotated probe is two uniform forms (`AnnotatedBindingProbeFormRequest` = wrappers only, `PlainAnnotatedBindingProbeFormRequest` = primitives only) instead of one mixed request. Mixing wrappers and primitives in a single `[FromForm]` model triggers a hybrid Swashbuckle shape (`allOf:[wrapper1, wrapper2, {primitives}]`) that's painful to navigate from a shared test.

Suite: 2,182 tests green (was 2,170; +12 from the split form / primitive baselines).

PR #93, opened 2026-05-04
Add StrongTypes.AspNetCore: model binders for NonEmptyEnumerable<T> and Maybe<T>

### LINKED ISSUE #86: Bind NonEmptyEnumerable<T> and Maybe<T> from non-body sources
Follow-up to #75 / #81 / #84.

`IParsable<T>` covers every strong type whose wire form is a single string (`NonEmptyString`, `Email`, `Digit`, the four numeric wrappers). Two types still don't bind from `[FromQuery]` / `[FromRoute]` / `[FromHeader]` / `[FromForm]` because they don't fit that shape:

## `NonEmptyEnumerable<T>`

The wire form is *multiple* strings (`?xs=1&xs=2&xs=3`), not one. ASP.NET Core's collection binder produces `T[]`; turning that into `NonEmptyEnumerable<T>` requires a custom `IModelBinderProvider` that:

1. Recognizes `NonEmptyEnumerable<T>` parameter types.
2. Delegates to the framework's array binder for the elements.
3. Wraps the resulting array via `NonEmptyEnumerable.TryCreateRange`.
4. Reports a binding error (so `[ApiController]` returns 400 + `ValidationProblemDetails`) when the resulting array is empty.

## `Maybe<T>`

The natural mapping is *missing* → `None`, *present* → `Some(parse(s))`. `IParsable<T>.TryParse` only describes the second case — there's no string for "missing", so it can't distinguish present-empty from absent. A custom binder can read the binding context directly to detect absence.

In practice `T?` already covers the query-param use case more simply, so `Maybe<T>` binding is lower priority — but if we ship the binder package for `NonEmptyEnumerable<T>` it's natural to include.

## Where to ship it

The binder package needs an ASP.NET Core dependency, which the core `StrongTypes` package deliberately avoids. The right home is a new `Kalicz.StrongTypes.AspNetCore` project (the package shape called out by #75), shipping:

- `NonEmptyEnumerableModelBinderProvider`
- `MaybeModelBinderProvider`
- `IServiceCollection.AddStrongTypes()` extension that registers them in one call (`MvcOptions.ModelBinderProviders.Insert(0, …)`).

## Done when

- `StrongTypes.AspNetCore` package builds and ships.
- `BindingProbeController` exposes `NonEmptyEnumerable<int>` and (optionally) `Maybe<int>` parameters from query, route, header, and form.
- `NonBodyBindingTests` covers both round-trip and 400 / `ValidationProblemDetails` cases for each.
- `[FromBody]` round-trip already works via the existing JSON converter — confirm it still passes after adding the binder.

KaliCZ:
## Summary

Adds a niche companion package, `Kalicz.StrongTypes.AspNetCore`, containing the two MVC model binders that the framework can't synthesise from `IParsable<T>` alone.

- **`NonEmptyEnumerableModelBinder<T>`** — reads multiple raw strings from the binding source, parses each via `IParsable<T>`, wraps via `NonEmptyEnumerable.TryCreateRange`. Empty/missing source surfaces as 400 + `ValidationProblemDetails` (or `null` when the action parameter is `NonEmptyEnumerable<T>?`).
- **`MaybeModelBinder<T>`** — supports `Maybe<T>?` patch-contract semantics: omitted on a nullable target → `null` (skip), omitted on a non-nullable target → `default(Maybe<T>)`, present-but-empty → `None`, non-empty → `Some(parse(s))`. Parse failure → 400.
- **Single registration**: `services.AddStrongTypes()` inserts both providers at `MvcOptions.ModelBinderProviders[0]`.

## Element type support — `IParsable<T>`

The binders parse elements through a reflection-cached `StringElementParser<T>` that targets `IParsable<TSelf>.TryParse(string?, IFormatProvider?, out T)`. That covers every BCL primitive in net7+ **and** every `Kalicz.StrongTypes` wrapper, so `NonEmptyEnumerable<Positive<int>>` and `Maybe<NonEmptyString>` work out of the box and wire-level invariant violations (empty `NonEmptyString`, non-positive `Positive<int>`, malformed `Email`) surface as 400 + `ValidationProblemDetails` automatically.

Wrapper-of-wrapper combinations (`NonEmptyEnumerable<Maybe<T>>`, `Maybe<NonEmptyEnumerable<T>>`, …) are deliberately out of scope on non-body sources — there's no clean wire form. `[FromBody]` already handles arbitrary nesting through the JSON converters in the core package.

## Scope — when this package is needed

Only when a controller binds `Maybe<T>` or `NonEmptyEnumerable<T>` from `[FromForm]` (the primary use), `[FromQuery]`, `[FromHeader]`, or `[FromRoute]`. JSON APIs don't need it — `[FromBody]` already round-trips both via the core package's converters, and every other wrapper (`NonEmptyString`, `Email`, `Digit`, the numeric ones) binds from non-body sources via the framework's built-in `IParsable<T>` binder.

## Tests

- New `StrongTypes.AspNetCore.IntegrationTests` project + `StrongTypes.AspNetCore.TestApi` test host, with [`MaybeBindingTests.cs`](src/StrongTypes.AspNetCore.IntegrationTests/Tests/BindingTests/MaybeBindingTests.cs), [`NonEmptyEnumerableBindingTests.cs`](src/StrongTypes.AspNetCore.IntegrationTests/Tests/BindingTests/NonEmptyEnumerableBindingTests.cs), and a shared [`BindingTestAsserts.cs`](src/StrongTypes.AspNetCore.IntegrationTests/Tests/BindingTests/BindingTestAsserts.cs) covering round-trip, missing, empty, and invalid cases on each non-body source for both required and nullable target shapes.
- Existing `StrongTypes.Api` `BindingProbeController` extended with `Email`, `Digit`, and `Negative<decimal>` coverage; the corresponding integration tests were split per source into [`QueryBindingTests`](src/StrongTypes.Api.IntegrationTests/Tests/BindingTests/QueryBindingTests.cs), [`HeaderBindingTests`](src/StrongTypes.Api.IntegrationTests/Tests/BindingTests/HeaderBindingTests.cs), and [`FormBindingTests`](src/StrongTypes.Api.IntegrationTests/Tests/BindingTests/FormBindingTests.cs). Unsupported wrapper-of-wrapper cases get explicit `query-unsupported-*` / `header-unsupported-*` / `form-unsupported-*` probes.
- `[FromBody]` round-trip of both wrappers via the existing JSON converters is reconfirmed.

## Also in this PR — docs, skill, diagram

The original push focused on the binders; the rest of the PR is the documentation + tooling sweep that makes the package discoverable and correctly framed:

- **Package + repo readmes** ([src/StrongTypes.AspNetCore/readme.md](src/StrongTypes.AspNetCore/readme.md), [readme.md](readme.md)) reframed around the actual use case: forms first, JSON APIs don't need it, `Maybe<T>` from query/headers binds but the third state isn't useful there.
- **Claude / Codex skill**: new [`Skill/references/aspnetcore.md`](Skill/references/aspnetcore.md) and [`Skill/references/wpf.md`](Skill/references/wpf.md), with `Skill/SKILL.md` updated to list both packages and flag AspNetCore as niche / not for JSON APIs.
- **Skill-sync reminders** added to [`CLAUDE.md`](CLAUDE.md) and [`CONTRIBUTING.md`](CONTRIBUTING.md) so future package or wiring changes don't leave the skill stale.
- **Complete package-layout diagram redesign** ([docs/diagrams/package-layout.svg](docs/diagrams/package-layout.svg) + dark variant): four groups — Core, Web/API surface (Swashbuckle preferred, Microsoft tagged "not recommended", AspNetCore tagged "rarely needed" and visually de-emphasised), Utility packages (EfCore, FsCheck), App surface (WPF + GitHub-issue CTA inviting WinForms / MAUI / Avalonia requests).

Closes #86.

## Test plan

- [x] Full solution build clean with `TreatWarningsAsErrors`.
- [x] `StrongTypes.AspNetCore.IntegrationTests` — round-trip, missing, empty, invalid for both binders on every non-body source, both required and nullable target shapes.
- [x] `StrongTypes.Api.IntegrationTests` — `[FromBody]` round-trip of `NonEmptyEnumerable<T>` and `Maybe<T>` reconfirmed; per-source tests pass for the existing `IParsable<T>`-based wrappers.
- [x] `StrongTypes.Tests`, `StrongTypes.Analyzers.Tests`, `StrongTypes.OpenApi.IntegrationTests` — unrelated, unchanged, green.


src/StrongTypes.Api.IntegrationTests/Tests/BindingTests/NonEmptyEnumerableMaybeBindingTests.cs
```diff
@@ -0,0 +1,290 @@
...
+/// Verifies that <see cref="NonEmptyEnumerable{T}"/> and <see cref="Maybe{T}"/>
+/// bind from every non-body source via the model binders shipped in
+/// <c>Kalicz.StrongTypes.AspNetCore</c>, and that an empty
+/// <c>NonEmptyEnumerable</c> source produces a 400 with
+/// <c>ValidationProblemDetails</c>. Body round-trip is covered too — the
+/// existing JSON converters handle that path; we just confirm the binders
+/// don't break it.
+/// </summary>
+[Collection(IntegrationTestCollection.Name)]
+public sealed class NonEmptyEnumerableMaybeBindingTests(TestWebApplicationFactory factory) : IDisposable
```
KaliCZ: Not real case. Nobody is going to put Maybe into a NonEmptyEnumerable as a contract.

Make it a NonEmptyEnumerable of PositiveInt or NonEmptyString

src/StrongTypes.Api/Controllers/BindingProbeController.cs
```diff
@@ -83,6 +83,64 @@ public IActionResult FromForm([FromForm] BindingProbeFormRequest request)
             email = request.Email.Address,
             nullableEmail = request.NullableEmail?.Address,
         });
+
+    // ── NonEmptyEnumerable<int> + Maybe<int> via the StrongTypes.AspNetCore binders ──
+
+    [HttpGet("query-nee")]
+    public IActionResult NonEmptyEnumerableFromQuery(
+        [FromQuery] NonEmptyEnumerable<int> ids,
+        [FromQuery] Maybe<int> filter)
```
KaliCZ: Not real case to use Maybe in a Query. That's just a nullable int then. or does it differentiate the 3 states?

src/StrongTypes.Api/Controllers/BindingProbeController.cs
```diff
@@ -92,3 +150,11 @@ public sealed record BindingProbeFormRequest(
     Positive<int>? NullableCount,
     Email Email,
     Email? NullableEmail);
+
+public sealed record NonEmptyEnumerableFormRequest(
+    NonEmptyEnumerable<int> Ids,
+    Maybe<int> Filter);
```
KaliCZ: Rather than creating new request types, just extend the existing types with more properties. Then you'll realize that you also haven't tested the nullable versions - nullable maybe (which is really the only case for a Maybe in the contract - if it's not a nullable maybe, then it can just be T?) and nullable NonEmptyEnumerable.

src/StrongTypes.Api/Controllers/BindingProbeController.cs
```diff
@@ -92,3 +150,11 @@ public sealed record BindingProbeFormRequest(
...
+
```
KaliCZ, line 138: Also add tests for Digits

src/StrongTypes.AspNetCore/MaybeModelBinder.cs
```diff
@@ -0,0 +1,60 @@
+using System;
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Mvc.ModelBinding;
+
+namespace StrongTypes.AspNetCore;
+
+/// <summary>Binds an action parameter of type <see cref="Maybe{T}"/> from any non-body source. An absent source binds to <c>None</c>; a present source parses the raw string via <see cref="IParsable{TSelf}"/> on <typeparamref name="T"/> and wraps it as <c>Some</c>; a parse failure surfaces as a <c>400</c>.</summary>
+/// <typeparam name="T">The wrapped type. Must implement <see cref="IParsable{TSelf}"/> (every BCL primitive in net7+ and every Kalicz.StrongTypes wrapper qualifies).</typeparam>
+public sealed class MaybeModelBinder<T> : IModelBinder
```
KaliCZ, line 9: I'm thikning about whether this is even useful? Maybe is only going to be used to represent the 3 states - wipe vs skip. And I believe the binder genuinely skips the case when the value is not provided. Which means Maybe loses value completely.

PR #115, opened 2026-07-10
Support storing the Email strong type in EF Core

KaliCZ:
## What

Makes the `Email` strong type EF Core–storable. Previously only the BCL `MailAddress` could go in a database column; an `Email` property failed at model build (EF treated the wrapper as an owned type, "No suitable constructor for the type 'Email'"). Storing an email meant using `MailAddress` as an undocumented workaround — which the `Email` XML doc even contradicted by claiming EF-column support.

## Changes

- **`EmailValueConverter`** — round-trips `Email` ⇄ `string` via the address, re-validating on read (a stored value that violates the email contract throws rather than materialising a broken wrapper, consistent with the other converters).
- **Convention wiring** — `ResolveConverter`/`IsStrongType` now recognise `Email`, so the converter attaches automatically. No `HasConversion` by hand.
- **Both storage types tested** — the harness entity that stores `MailAddress` is renamed `EmailEntity` → `MailAddressEntity` (its name now matches its column), and a new `EmailEntity` stores the `Email` wrapper. Both run the full entity-test matrix (round-trip of value + null, `NULL`/value filtering, ordering, read-invalid) for their nullable forms.
- **Analyzer** — `MissingEfCorePackageAnalyzer` (ST0001) now flags `Email` properties on EF entities, like `NonEmptyString`. `MailAddress` stays out — it's a BCL type, not a StrongTypes wrapper.
- **Docs** — `efcore.md` and the EfCore readme list `Email`/`MailAddress` as storable, and note the two are interchangeable (store whichever your domain holds). The `Email` `<remarks>` claim about EF columns is now true.

## Verification

- Full solution build: 0 warnings, 0 errors.
- Full `StrongTypes.Api.IntegrationTests` on PostgreSQL: **1070 passed, 17 skipped, 0 failed** (the 17 are SQL-Server rows under the local opt-out; CI runs both providers).
  - New `EmailEntityTests` (stores `Email`): 37 passed.
  - `MailAddressEntityTests` + `MailAddressFilterTests` (renamed): 41 passed / 4 skipped.
- Analyzer tests: 12 passed (incl. the new `Detects_email_wrapper`).

### DIFF
--- a/Skill/references/efcore.md
+++ b/Skill/references/efcore.md
@@ -30,14 +30,18 @@ public sealed class User
     public Guid Id { get; init; }
     public NonEmptyString Name { get; init; }
     public NonEmptyString? Nickname { get; init; }
+    public Email? ContactEmail { get; init; }
     public Positive<int>   LoginCount { get; init; }
     public NonNegative<decimal> Balance { get; init; }
 }
 ```
 
-Columns are the underlying type — `nvarchar` for `NonEmptyString`, `int`
-for `Positive<int>`, `decimal(...)` for `NonNegative<decimal>`, and the
-nullable form becomes a nullable column.
+Columns are the underlying type — `nvarchar` for `NonEmptyString` and for
+`Email` / `MailAddress` (the address string), `int` for `Positive<int>`,
+`decimal(...)` for `NonNegative<decimal>`, and the nullable form becomes a
+nullable column. An email column can be `Email` (the wrapper — re-checks the
+254-char cap on read) or the BCL `MailAddress`; they convert to each other
+implicitly, so store whichever your domain already holds.
 
 ### Non-public backing properties
 
--- a/Skill/references/efcore.md
+++ b/Skill/references/efcore.md
@@ -31,6 +31,7 @@ public sealed class User
     public NonEmptyString Name { get; init; }
     public NonEmptyString? Nickname { get; init; }
     public Email? ContactEmail { get; init; }
+    public MailAddress? BackupEmail { get; init; }
     public Positive<int>   LoginCount { get; init; }
     public NonNegative<decimal> Balance { get; init; }
 }
--- a/src/StrongTypes.Api.IntegrationTests/Tests/ApiTests/Emails/EmailEntityTests.cs
+++ b/src/StrongTypes.Api.IntegrationTests/Tests/ApiTests/Emails/EmailEntityTests.cs
@@ -14,11 +14,6 @@ public sealed class EmailEntityTests(TestWebApplicationFactory factory)
     protected override string FirstValid => "alice@example.com";
     protected override string UpdatedValid => "bob@example.org";
 
-    // Email isn't IComparable; the column stores the address string, so the
-    // database orders by that. Mirror it here for the OrderBy assertion.
-    protected override IComparer<Email> ValueComparer =>
-        Comparer<Email>.Create((left, right) => string.CompareOrdinal(left.Address, right.Address));
-
     public static TheoryData<string> ValidInputs => new()
     {
         "alice@example.com",
--- a/src/StrongTypes.EfCore/StrongTypesConvention.cs
+++ b/src/StrongTypes.EfCore/StrongTypesConvention.cs
@@ -60,14 +60,15 @@ internal sealed class StrongTypesConvention : IEntityTypeAddedConvention, IPrope
     private static ValueConverter? ResolveConverter(Type clrType)
     {
         var unwrapped = Nullable.GetUnderlyingType(clrType) ?? clrType;
-        return unwrapped switch
-        {
-            _ when unwrapped == typeof(NonEmptyString) => new NonEmptyStringValueConverter(),
-            _ when unwrapped == typeof(Email) => new EmailValueConverter(),
-            _ when unwrapped == typeof(MailAddress) => new MailAddressValueConverter(),
-            { IsGenericType: true } when IsNumericWrapper(unwrapped) => CreateNumericConverter(unwrapped),
-            _ => null,
-        };
+        if (unwrapped == typeof(NonEmptyString))
+            return new NonEmptyStringValueConverter();
+        if (unwrapped == typeof(Email))
+            return new EmailValueConverter();
+        if (unwrapped == typeof(MailAddress))
+            return new MailAddressValueConverter();
+        if (unwrapped.IsGenericType && IsNumericWrapper(unwrapped))
+            return CreateNumericConverter(unwrapped);
+        return null;
     }
 
     private static bool IsNumericWrapper(Type unwrapped)
--- a/src/StrongTypes.EfCore/StrongTypesConvention.cs
+++ b/src/StrongTypes.EfCore/StrongTypesConvention.cs
@@ -52,23 +52,7 @@ internal sealed class StrongTypesConvention : IEntityTypeAddedConvention, IPrope
         }
     }
 
-    private static bool IsStrongType(Type clrType)
-    {
-        var unwrapped = Nullable.GetUnderlyingType(clrType) ?? clrType;
-        if (unwrapped == typeof(NonEmptyString) || unwrapped == typeof(Email) || unwrapped == typeof(MailAddress))
-        {
-            return true;
-        }
-        if (unwrapped.IsGenericType)
-        {
-            var definition = unwrapped.GetGenericTypeDefinition();
-            return definition == typeof(Positive<>)
-                || definition == typeof(NonNegative<>)
-                || definition == typeof(Negative<>)
-                || definition == typeof(NonPositive<>);
-        }
-        return false;
-    }
+    private static bool IsStrongType(Type clrType) => ResolveConverter(clrType) is not null;
 
     // EF Core applies ValueConverter only to non-null values, so the same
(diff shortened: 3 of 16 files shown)


Skill/references/efcore.md
```diff
@@ -30,14 +30,18 @@ public sealed class User
     public Guid Id { get; init; }
     public NonEmptyString Name { get; init; }
     public NonEmptyString? Nickname { get; init; }
+    public Email? ContactEmail { get; init; }
```
KaliCZ, line 33: Also show the MailAddress

src/StrongTypes.Api.IntegrationTests/Tests/ApiTests/Emails/EmailEntityTests.cs
```diff
@@ -7,18 +6,18 @@ namespace StrongTypes.Api.IntegrationTests.Tests;
...
+    protected override Email Create(string raw) => Email.Create(raw);
     protected override string FirstValid => "alice@example.com";
     protected override string UpdatedValid => "bob@example.org";
 
-    // MailAddress isn't IComparable; the column stores the address string, so the
+    // Email isn't IComparable; the column stores the address string, so the
     // database orders by that. Mirror it here for the OrderBy assertion.
-    protected override IComparer<MailAddress> ValueComparer =>
-        Comparer<MailAddress>.Create((left, right) => string.CompareOrdinal(left.Address, right.Address));
+    protected override IComparer<Email> ValueComparer =>
```
KaliCZ: We shouldn't need this override here anymore, right?

src/StrongTypes.EfCore/readme.md
```diff
@@ -49,6 +50,7 @@ public sealed class User
     public Guid Id { get; init; }
     public NonEmptyString Name { get; init; }
     public NonEmptyString? Nickname { get; init; }
+    public Email? ContactEmail { get; init; }
```
KaliCZ, line 53: Also show the MailAddress

src/StrongTypes.EfCore/StrongTypesConvention.cs
```diff
@@ -55,7 +55,7 @@ public void ProcessPropertyAdded(
     private static bool IsStrongType(Type clrType)
     {
         var unwrapped = Nullable.GetUnderlyingType(clrType) ?? clrType;
-        if (unwrapped == typeof(NonEmptyString) || unwrapped == typeof(MailAddress))
+        if (unwrapped == typeof(NonEmptyString) || unwrapped == typeof(Email) || unwrapped == typeof(MailAddress))
```
KaliCZ: using the `is ... or` would be much shorter, right?

src/StrongTypes.EfCore/StrongTypesConvention.cs
```diff
@@ -80,6 +80,10 @@ private static bool IsStrongType(Type clrType)
         {
             return new NonEmptyStringValueConverter();
         }
+        if (unwrapped == typeof(Email))
+        {
+            return new EmailValueConverter();
+        }
         if (unwrapped == typeof(MailAddress))
```
KaliCZ: Can't this be a match?

Skill/references/efcore.md
KaliCZ: Done in 5658ae3 — added a `MailAddress? BackupEmail` property alongside `Email? ContactEmail` in the example.

src/StrongTypes.EfCore/readme.md
KaliCZ: Done in 5658ae3 — added `MailAddress? BackupEmail` to the example.

src/StrongTypes.Api.IntegrationTests/Tests/ApiTests/Emails/EmailEntityTests.cs
KaliCZ: You're right that it shouldn't be needed — the only reason it was is that `Email` wasn't `IComparable`, so `Comparer<Email>.Default` (what the base `ValueComparer` defaults to) throws at runtime. Fixed the root cause in 5658ae3: `Email` now implements `IComparable<Email>` — ordinal-ignore-case on the address, matching its `Equals` so `CompareTo == 0` agrees with equality. The override is gone and the base default just works, and I added unit tests for the ordering.

(`MailAddressEntityTests` keeps its override — `MailAddress` is a BCL type we can't make `IComparable`.)

src/StrongTypes.EfCore/StrongTypesConvention.cs
KaliCZ: `is`/`or` patterns don't work on `Type` equality here — `x is typeof(NonEmptyString)` checks `x`'s runtime type (always `Type`), and `typeof(...)` isn't a constant pattern you can put after `is` / `or`. But the whole method was just repeating `ResolveConverter`'s type list, so in 5658ae3 I collapsed it to `IsStrongType(clrType) => ResolveConverter(clrType) is not null` — shorter than either form, and no duplicated list to keep in sync.

src/StrongTypes.EfCore/StrongTypesConvention.cs
KaliCZ: Done in 5658ae3 — `ResolveConverter` is a `switch` expression now. Type dispatch still needs `when` guards (you can't `case typeof(...)`), but it reads as a match, with the numeric-wrapper detection and construction factored into small `IsNumericWrapper` / `CreateNumericConverter` helpers.

src/StrongTypes.EfCore/StrongTypesConvention.cs
```diff
@@ -52,56 +52,38 @@ public void ProcessPropertyAdded(
...
-                var underlying = unwrapped.GetGenericArguments()[0];
-                var converterType = typeof(NumericStrongTypeValueConverter<,>).MakeGenericType(unwrapped, underlying);
-                return (ValueConverter)Activator.CreateInstance(converterType)!;
-            }
-        }
-        return null;
+            _ when unwrapped == typeof(NonEmptyString) => new NonEmptyStringValueConverter(),
+            _ when unwrapped == typeof(Email) => new EmailValueConverter(),
+            _ when unwrapped == typeof(MailAddress) => new MailAddressValueConverter(),
+            { IsGenericType: true } when IsNumericWrapper(unwrapped) => CreateNumericConverter(unwrapped),
```
KaliCZ: I guess rather than hacking, let's just use ifs. But we dont need brackets. So we can do something like

if (unwrapped == typeof(NonEmptyString))
    return new NonEmptyStringValueConverter();
if (unwrapped == typeof(Email))
    return new EmailValueConverter();
if (unwrapped == typeof(MailAddress))
    return new MailAddressValueConverter();
if (unwrapped.IsGenericType && IsNumericWrapper(unwrapped))
    return something

src/StrongTypes.EfCore/StrongTypesConvention.cs
```diff
@@ -52,56 +52,38 @@ public void ProcessPropertyAdded(
         }
     }
 
-    private static bool IsStrongType(Type clrType)
```
KaliCZ, line 55: Don't we have similar types also used in the OpenAPI? We could maybe abstract such logic to a common shared library if useful

src/StrongTypes.EfCore/StrongTypesConvention.cs
KaliCZ: Done in 576afdf — swapped the switch for the bracket-less if-ladder exactly as you laid out, `return CreateNumericConverter(unwrapped)` as the last branch. Agree it's clearer; the `_ when` guards were only there because `Type` equality can't `case` cleanly, and plain ifs sidestep that entirely.

src/StrongTypes.EfCore/StrongTypesConvention.cs
KaliCZ: Good eye — the set is enumerated in `StrongTypes.EfCore` (this file), `StrongTypes.OpenApi.Core` (`StrongTypeSchemaTypes`), `StrongTypes.Wpf`, and the analyzer. I dug in, and I don't think a shared list buys as much as it looks, for two reasons:

1. **The subsets differ per consumer.** `StrongTypeSchemaTypes` covers `Digit`, `Maybe<>`, and `NonEmptyEnumerable<>` but *not* `MailAddress`; EfCore covers `MailAddress` but not those (they aren't storable scalars). So there's no single shared list — each package enumerates a different slice.

2. **The per-type payload is entirely different, and that's the valuable part.** EfCore maps type → `ValueConverter`; OpenAPI maps type → wire type + schema bounds; the analyzer is `netstandard2.0` and works on Roslyn *symbols* / metadata-name strings (`"StrongTypes.Positive\`1"`) — it can't reference the runtime `Type`s at all. Only the thin plumbing (`UnwrapNullable`, "is this generic definition") is genuinely common.

So a shared catalog would centralize "these type identities exist" but each consumer still writes its own behavior, and the analyzer stays separate no matter what. Real duplication, modest payoff, and it's cross-cutting across 4+ packages — I'd keep it out of this PR and treat it as its own spike. Happy to file a task to prototype a core `StrongTypeCatalog` if you want to see how thin/thick it lands.

